PR #10, opened 2026-07-16
fix: stop using kqueue watch backend on macOS, canonicalize watch roots

### LINKED ISSUE #9: macOS: MCP worker leaks file descriptors until system-wide kern.maxfiles exhaustion (kqueue watchers: 1 FD/file, leaked across restarts, orphaned --ui workers never exit)
### Description

On macOS, long-running `infigraph-mcp` worker processes accumulate open file descriptors without bound. In a real working session, two worker processes reached **176,463 and 183,252 open FDs** — together ~97% of the entire OS file table (`kern.num_files` 368,616 vs. `kern.maxfiles` 368,640) — at which point unrelated shell commands (`echo hi`, `ulimit -n`) began failing with *"too many open files in system"*. Killing the two workers dropped `kern.num_files` to 8,877 within seconds. This recurred twice in one session over several hours; recovery required externally killing the workers each time.

The leak is a compound of five defects (details in *Root cause* below):

1. The `notify` kqueue backend is enabled for the **whole binary** via Cargo feature unification, so code watchers hold **one open FD per watched file and per directory** instead of using FSEvents.
2. Watchers register **every directory in the tree individually** and, on kqueue, every file — a single watcher on a 315-file repo measurably holds **~1,660 FDs**.
3. The watcher **restart path leaks each previous generation's FDs** — directories observed held 4× (= initial + `MAX_RESTARTS = 3`) by a process whose watcher registry listed *no* watcher for those paths.
4. `auto_start_watch_opportunistic` **bypasses the `watchers_disabled` guard** that the instance lock sets for non-primary workers, and there is no cross-process dedup — so every worker (including supposedly watcher-disabled secondaries) starts full watcher sets on `index_project`, `group_*`, and even `search` calls.
5. In `--ui` mode the worker **never exits after stdin EOF** (infinite sleep loop), so every MCP client disconnect strands an immortal worker holding all of the above. Orphans accumulate: four launcher+worker pairs, all bound to `--port=9749`, were observed alive simultaneously within minutes of normal use.
6. The same leak also manifests in **detached `infigraph watch` CLI processes**: `ensure_watcher_running` (`crates/infigraph-cli/src/index.rs:275`) spawns one per project on every CLI `index` — including when the MCP server shells out to the CLI in `tool_index_project` (`crates/infigraph-mcp/src/tools/index.rs`) — and one such process was observed at **184,324 FDs ~25 minutes after spawn** while watching a single large repo; killing it dropped `kern.num_files` from 199,277 to 14,916.

### Environment

- macOS 26.5.2 (Darwin 25.5.0), arm64 (Apple Silicon)
- `infigraph-mcp` invoked as `infigraph-mcp --ui --mcp --port=9749` (supervisor self-spawns `--worker`)
- Multiple projects registered (~15), including large/nested/overlapping roots

### Evidence

Observed directly on a live instance (independent observation, before any code reading):

```
# Two stale workers, both --worker --ui --mcp --port=9749:
lsof -p <workerA> | wc -l   → 176,463
lsof -p <workerB> | wc -l   → 183,252
# Their sibling (supervisor) launcher processes: 7 FDs each.

sysctl kern.num_files       → 368,616
sysctl kern.maxfiles        → 368,640   (~97% of the OS file table held by 2 processes)

# System-wide failures: `echo hi`, `ulimit -n` → "too many open files in system"
kill <workerA> <workerB>    → kern.num_files drops to 8,877 within seconds
```

Reproduced fresh (same machine, minutes after restart):

```
# FD type breakdown of one accumulating worker (lsof -p <pid>, column TYPE):
4827 DIR      ← open directory FDs under watched project trees
  27 KQUEUE   ← 27 kqueue instances in ONE process (registry listed only 2 watchers)
   8 REG, 3 PIPE, 1 IPv4

# Of the 4,827 DIR FDs, only 2,465 unique paths — many directories held exactly 4×
# (= 1 initial + MAX_RESTARTS(3) leaked watcher generations).
# The process's own watcher registry (get_watch_status) listed NO watcher for those paths.

# Cost of a single fresh watcher on a small repo (this repo, 315 source files, cloned):
#   1,063 REG (one per file) + 598 DIR (one per directory) ≈ 1,660 FDs for one watcher.

# Orphan accumulation: 4 launcher+worker pairs alive simultaneously,
# all `--ui --mcp --port=9749`, started 12:49 / 12:51 / 12:53 / 12:55.

# <project>/.infigraph/watch.log from the earlier incident is filled end-to-end with:
watch error: Too many open files (os error 24)

# Control: a CLEAN stop does release everything — calling stop_watch + stop_watch_docs
# on the fresh worker dropped it from 1,675 FDs to 7 within seconds.
# So FDs are only lost when watchers are never stopped (orphaned workers, duplicate
# watcher sets) or when the watcher dies and is restarted in-process.
```

### Repro steps

1. macOS, any project. Configure `infigraph-mcp --ui --mcp --port=9749` as a stdio MCP server (e.g. in Claude Code).
2. Start a session; call `index_project` (auto-watch starts). `lsof -p $(pgrep -f 'infigraph-mcp --worker') | wc -l` → roughly *(files + dirs)* FDs, plus the same again for the doc watcher.
3. End the session (stdin EOF). The launcher+worker pair stays alive forever (`main.rs:360-365`).
4. Start a new session; index/search again. A new pair spawns; the new worker starts its own watchers even though the stale worker holds `~/.infigraph/mcp.lock` (the opportunistic path skips the disabled check). Repeat across sessions/hours.
5. To see the restart amplification: delete/recreate directories inside a watched tree (e.g. a build writing into a non-ignored dir) so the kqueue watcher's internal thread dies; each auto-restart re-opens the full tree while the dead generation's FDs remain (duplicate DIR entries in `lsof`, up to 4×).
6. `sysctl kern.num_files` climbs until system-wide EMFILE.

### Suspected root cause

**(1) kqueue backend enabled binary-wide by feature unification.**
`crates/infigraph-docs/Cargo.toml:29` enables it for doc watching:

```toml
notify = { version = "6", features = ["macos_kqueue"] }
```

while `crates/infigraph-core/Cargo.toml:33-34` deliberately does *not*:

```toml
# macos_kqueue enables the kqueue backend on macOS; notify selects inotify on Linux automatically.
notify = "6"
```

Cargo unifies features per crate version across the build graph, so the `infigraph-mcp` binary compiles `notify` **with** `macos_kqueue`, and `RecommendedWatcher` in `infigraph-core` becomes the kqueue backend too — defeating core's opt-out (notify's cfg gates give `macos_kqueue` precedence over the default `macos_fsevent` when both are enabled). Confirmed in the resolved graph:

```
$ cargo tree -e features -i notify -p notify
notify v6.1.1
├── notify feature "kqueue"
│   └── notify feature "macos_kqueue"
│       └── infigraph-docs v2.8.0 (crates/infigraph-docs)   ← sole enabler
├── notify feature "macos_fsevent" (*)
├── notify feature "macos_kqueue" (*)
```

kqueue needs an open FD per watched vnode (every file + every directory); FSEvents needs a handful per stream. Empirically: 1,063 REG + 598 DIR FDs for one watcher over a 315-file repo, plus `KQUEUE`-type FDs visible in `lsof` at runtime.

**(2) Per-directory registration amplifies the FD count.**
`register_watch_dirs` / `register_subdirs` (`crates/infigraph-core/src/watch/mod.rs:498-526`) walk the tree and `watch()` every directory `NonRecursive`ly (to skip ignore-dirs), and new dirs are added on Create events (`watch/mod.rs`, `register_subdirs` call in the event loop). Fine on FSEvents/inotify; on kqueue it pins the whole tree open.

**(3) Watcher restart leaks previous generations.**
In `watch_project_with_periodic` (`crates/infigraph-core/src/watch/mod.rs:80-296`), when the notify channel disconnects ("Watcher's internal thread died (e.g. kqueue panic on dir deletion)" — the code's own comment), `create_watcher` builds a fresh watcher registering the full tree again, up to `MAX_RESTARTS = 3`. Observed state — directories held exactly 4× by a process whose watcher registry listed no watcher on those paths — shows the dropped watchers' FDs are not reclaimed after the backend's internal thread has died. (A *clean* stop does reclaim: `stop_watch` + `stop_watch_docs` dropped a live worker from 1,675 FDs to 7 — the leak is specific to the death/restart path and to watchers that are simply never stopped.) After the final "giving up", the thread exits and the registry entry is removed (`tools/watch.rs` end of `tool_watch_project` thread), but all 4 generations of FDs remain until process exit.

**(4) "Watchers disabled" is bypassed; no cross-process dedup.**
Non-primary workers (instance lock `~/.infigraph/mcp.lock` held elsewhere) call `disable_watchers()` (`crates/infigraph-mcp/src/main.rs`, start of `run()`), but the auto-start actually used everywhere is the bypass variant:

- `crates/infigraph-mcp/src/tools/watch.rs:71-73` — `auto_start_watch_opportunistic` → `auto_start_watch_inner(path, /* skip_disabled_check = */ true)`
- imported as `auto_start_watch` in `tools/index.rs:8` and `tools/groups.rs:10`; also called directly from `tools/search.rs:517`

So `index_project`, `scip_import`, `group_index`, and even `search` start watchers in *every* worker. Dedup (`is_watching`, `tools/watch.rs:60-65`) is an in-process map only; `cli_watcher_holds_lock` (`tools/watch.rs:107-126`) checks only the CLI's per-project `watch.lock`, which MCP watchers never acquire — so N workers = N duplicate watcher sets over the same trees.

**(5) `--ui` workers are immortal after client disconnect.**
`run()` (`crates/infigraph-mcp/src/main.rs:360-365`):

```rust
// If UI mode is active, keep process alive after stdin EOF (web server still serving)
if ui_enabled {
    loop {
        std::thread::sleep(std::time::Duration::from_secs(3600));
    }
}
```

Every MCP client disconnect leaves the worker (and its supervisor parent, blocked in `cmd.status()`, `main.rs:9-66`) running forever with all watcher FDs. The next session spawns a new pair (UI port bind fails gracefully: "port already in use — skipping UI"), so pairs — and their FD sets — accumulate monotonically. This is also why two workers on the same `--port` are observed at once.

**Combined arithmetic:** (files+dirs per tree) × (code + doc watcher) × (up to 4 leaked generations) × (N orphaned workers), across ~15 registered projects with overlapping roots → the observed ~180k FDs per worker.

### Why kqueue was probably chosen — and why FSEvents only *looks* broken

With `macos_kqueue` removed, watchers initially appear dead: no events are ever delivered to the CLI watcher. The actual bug is that `cmd_watch` (`crates/infigraph-cli/src/info_commands.rs:341-365`) watches its root as the **relative path `.`** ("Watching . (debounce 500ms)"). FSEvents delivers absolute, `realpath`ed event paths, so both notify's internal path filtering and the event loop's `strip_prefix(root)` (`watch/mod.rs`, `Err(_) => continue`) silently drop every event when the root is relative. kqueue is FD-based and returns paths as registered, so it masked the bug. Canonicalizing the root fixes FSEvents completely — verified on a patched build:

| | kqueue build (v2.7.1 release) | FSEvents build (`macos_kqueue` removed + root canonicalized) |
|---|---|---|
| watcher on 40-file / 20-dir project | 83 FDs (73 pinned to the tree) | **8–9 FDs (0 pinned)** |
| watcher on one large repo | 184,324 FDs (observed live) | O(1) |
| detects modify/create, reindexes | yes | **yes** (verified: `[watch] batch indexing 1 files`) |

(The MCP-side `tool_watch_project` already canonicalizes its root, so it needs no change to work on FSEvents.)

### Suggested fix

In rough priority order:

1. **Stop using kqueue on macOS for tree watching.** Remove `macos_kqueue` from `infigraph-docs/Cargo.toml:29` (because of feature unification this single line currently opts the whole binary in), and canonicalize the watch root in `cmd_watch` (`infigraph-cli/src/info_commands.rs:341`) so FSEvents event filtering works (see previous section — verified fix, FD count per watcher drops from O(files+dirs) to O(1) with change detection intact). Optionally also switch `register_watch_dirs` to a single `watch(root, RecursiveMode::Recursive)` + the existing event-side `should_ignore` filtering, since per-directory registration is only needed for kqueue/inotify-style backends.
2. **Exit the worker when the client disconnects.** After stdin EOF, shut down watchers and exit (or add an idle timeout / exit when no UI+MCP activity for N minutes). At minimum, stop all watchers before parking in the sleep loop.
3. **Honor `watchers_disabled` in the opportunistic path**, or replace the in-process dedup with a cross-process per-project lock (extend the existing `.infigraph/watch.lock` mechanism to MCP watchers) so N workers can't hold N duplicate watcher sets.
4. **Fix the restart path** to guarantee the old watcher's FDs are released before re-registering (explicitly `unwatch`/drop and verify closure; the notify kqueue backend does not appear to release per-vnode FDs once its internal thread has panicked). Consider failing over to a fresh process instead of in-process re-registration.
5. Surface FD/watcher diagnostics in `get_watch_status` (kqueue instance count, FDs held) so this is observable before the OS file table fills.

### Workaround (for anyone hitting this)

`pkill -f 'infigraph-mcp'` and `pkill -f 'infigraph watch'` reclaim the FDs immediately; the OS file table recovers within seconds (verified: killing one detached watcher dropped `kern.num_files` from 199,277 to 14,916). Stale processes re-accumulate with normal use, so this needs repeating until fixed.

### Existing-issue check

Searched issues/PRs for "file descriptor", "fd leak", "too many open files", "leak", "worker", "watch EMFILE" — no existing issue covers this (tracker has 3 closed issues: #6–#8; #6 is watcher-related but about reindexing behavior, not FDs).

pradeepmouli:
## Summary

Fixes #9. On macOS, `infigraph-mcp` workers leak file descriptors until `kern.maxfiles` is exhausted system-wide (observed: two workers holding 176k/183k FDs, ~97% of the OS file table).

Root cause, confirmed via `cargo tree -e features -i notify -p notify`: `infigraph-docs/Cargo.toml` enables notify's `macos_kqueue` feature for its own doc watcher. Cargo feature unification applies that feature workspace-wide to every crate depending on the same resolved `notify` version, so `infigraph-core`'s code watcher silently gets the kqueue backend too (its own `Cargo.toml` requests default features only — the opt-out is defeated). kqueue holds one open FD per watched file *and* per directory; a single watcher on a modest repo measurably held ~1,660 FDs, and on a real project tree that's tens of thousands, held open indefinitely.

This PR:
1. Removes `macos_kqueue` from `infigraph-docs/Cargo.toml`, switching macOS back to the default FSEvents backend (O(1) FDs regardless of tree size).
2. Canonicalizes `root` once at the top of `watch_project_with_periodic` (`infigraph-core/src/watch/mod.rs`). FSEvents delivers absolute, symlink-resolved event paths, so a non-canonical `root` (relative path, or one that traverses a symlink — macOS temp/build dirs commonly live under `/var`, itself a symlink to `/private/var`) made `path.strip_prefix(root)` fail silently for every event, dropping all changes with no error. This is almost certainly why kqueue was reached for in the first place — FSEvents looked broken when the actual bug was an uncanonicalized root. Fixing it in the core function (rather than at each call site) protects every caller — CLI `watch`, the MCP `watch_project` tool, and the doc watcher — at once.

## Test plan

- [x] `cargo test -p infigraph-core -p infigraph-docs` passes (single-threaded; parallel runs on this machine hit an unrelated Kuzu mmap resource-contention issue building N test DBs concurrently, not caused by this change — reproduces the same way on `main`)
- [x] `cargo clippy -p infigraph-core -p infigraph-docs --all-targets` passes with zero new warnings (one pre-existing `useless_borrows_in_formatting` lint in `vuln/mod.rs` reproduces identically on unpatched `main` with this toolchain — unrelated to this change, not touched here)
- [x] `cargo fmt --all -- --check` introduces zero new diffs (confirmed by diffing before/after this patch — pre-existing drift on ~76 unrelated files exists on `main` already with this local rustfmt version, not touched here)
- [x] Tested manually:
  - Added a regression test (`watch_project_detects_changes_through_symlinked_root`) that reproduces the symlink-root bug deterministically via a custom-prefixed `TempDir` and asserts a file-removal event is delivered. Verified it fails without the `canonicalize()` fix and passes with it.
  - Built both `infigraph` and `infigraph-mcp` release binaries with this patch and drove them directly over stdio (JSON-RPC) against a real 40-file test project: watcher FD count stayed at 8–9 for the life of the process (vs. 83 FDs, 73 pinned to the tree, on the unpatched kqueue build), and file modifications were correctly detected and triggered reindexing (`[watch] batch indexing 1 files`).
  - Found and killed a live, unpatched `infigraph watch` CLI process during this investigation holding **184,324 FDs** ~25 minutes after spawn; killing it dropped `kern.num_files` from 199,277 to 14,916 within seconds — confirming the FDs are genuinely held by the process (not a kernel accounting artifact) and are fully reclaimed on exit.

## Notes

- Full root-cause writeup, evidence, and repro steps are in #9.
- I considered fixing only the CLI's `cmd_watch` (which is the one call site that watches an unresolved `.`), but moving the canonicalize into `watch_project_with_periodic` itself is more robust — it can't be reintroduced by a future caller that forgets to canonicalize (the MCP tool's `tool_watch_project` already canonicalizes independently, so this is technically redundant for that path today, but removes the foot-gun for future call sites).
- Not addressed here (tracked as follow-ups in #9, since they're independent of the kqueue root cause): `--ui` workers never exiting after stdin EOF, `auto_start_watch_opportunistic` bypassing the `watchers_disabled` instance-lock guard, and the watcher-restart path leaking FDs from prior generations when the backend's internal thread dies. Happy to take these on in follow-up PRs if that's useful — wanted to keep this one scoped to the single root cause with a clean, verifiable diff.

### DIFF
--- a/crates/infigraph-core/src/watch/mod.rs
+++ b/crates/infigraph-core/src/watch/mod.rs
@@ -90,6 +90,13 @@ where
     MR: Fn() -> Result<crate::lang::LanguageRegistry> + Send + 'static,
     F: Fn(&crate::IndexResult) + Send + 'static,
 {
+    // Some watch backends (e.g. FSEvents on macOS) deliver absolute,
+    // symlink-resolved event paths regardless of how `root` was specified.
+    // If `root` is relative, or traverses a symlink (macOS temp dirs live
+    // under /var, itself a symlink to /private/var), `path.strip_prefix(root)`
+    // below silently fails for every event and all changes are dropped.
+    let root = &root.canonicalize().unwrap_or_else(|_| root.to_path_buf());
+
     let ignore_dirs: &[&str] = &[
         ".infigraph",
         ".git",
@@ -522,5 +529,85 @@ fn register_subdirs(watcher: &mut RecommendedWatcher, dir: &Path, ignore_dirs: &
         }
         let _ = watcher.watch(&path, RecursiveMode::NonRecursive);
         register_subdirs(watcher, &path, ignore_dirs);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::sync::Mutex;
+
+    /// Regression test for a macOS-specific bug: `watch_project_with_periodic`
+    /// used to compare raw filesystem-watch event paths against the caller's
+    /// `root` exactly as given. FSEvents delivers absolute, symlink-resolved
+    /// event paths, so a non-canonical `root` (e.g. a relative path, or one
+    /// that traverses a symlink) made `path.strip_prefix(root)` fail for
+    /// every event, silently dropping all changes with no error. This is the
+    /// same class of bug that made the `infigraph watch` CLI command — which
+    /// watched the unresolved `.` — appear to receive no events at all,
+    /// prompting a workaround (the kqueue backend) that caused a much larger
+    /// file-descriptor leak.
+    ///
+    /// A custom-prefixed `tempfile::TempDir` reproduces a non-canonical root
+    /// deterministically on macOS: it lives under `/var/folders/...`, itself
+    /// a symlink to `/private/var/folders/...`. (The default `TempDir::new()`
+    /// prefix starts with a dot, which the watcher's own hidden-file filter
+    /// would ignore regardless of this bug, so a custom prefix is used to
+    /// keep the test isolated to the canonicalization behavior.)
+    #[test]
+    #[cfg(target_os = "macos")]
+    fn watch_project_detects_changes_through_symlinked_root() {
+        let tmp = tempfile::Builder::new()
+            .prefix("infigraph-watch-test-")
+            .tempdir()
+            .unwrap();
+        let raw_root = tmp.path().to_path_buf();
+        let canonical_root = raw_root.canonicalize().unwrap();
+        assert_ne!(
+            raw_root, canonical_root,
+            "test assumption broken: TempDir root is already canonical on this machine"
+        );
+
+        let file_path = raw_root.join("watched.txt");
+        std::fs::write(&file_path, "v1").unwrap();
+
+        let events: Arc<Mutex<Vec<WatchEvent>>> = Arc::new(Mutex::new(Vec::new()));
+        let events_clone = Arc::clone(&events);
+        let (stop_tx, stop_rx) = mpsc::channel();
+
+        let handle = std::thread::spawn(move || {
+            watch_project(
+                &raw_root,
+                || Ok(crate::lang::LanguageRegistry::new()),
+                50,
+                stop_rx,
+                move |evt| events_clone.lock().unwrap().push(evt),
+            )
+        });
+
+        // Give the watcher time to register before triggering a change.
+        std::thread::sleep(Duration::from_millis(300));
+        std::fs::remove_file(&file_path).unwrap();
+
+        // Poll rather than a single fixed sleep: fast on a quiet machine,
+        // robust on a loaded one.
+        let mut seen = false;
+        for _ in 0..40 {
+            std::thread::sleep(Duration::from_millis(100));
+            if !events.lock().unwrap().is_empty() {
+                seen = true;
+                break;
+            }
+        }
+
+        let _ = stop_tx.send(());
+        let _ = handle.join();
+
+        assert!(
+            seen,
+            "watch_project delivered no events for a change under a non-canonical \
+             (symlinked) root — the root.canonicalize() call in \
+             watch_project_with_periodic may have regressed"
+        );
     }
 }
--- a/crates/infigraph-docs/Cargo.toml
+++ b/crates/infigraph-docs/Cargo.toml
@@ -26,7 +26,7 @@ pdf_oxide = "0.3"
 calamine = "0.35"
 zip = "2"
 quick-xml = "0.37"
-notify = { version = "6", features = ["macos_kqueue"] }
+notify = "6"
 chrono = "0.4.45"
 fs2 = "0.4.3"
 


johnintuit (approved):

johnintuit:
Looks good and partially fix 

Issue #9 lists 5 defects. Only 2 are actually fixed across PR #10 + PR #13:
- ✅ Defect 1 (kqueue feature unification) — PR #10
- ✅ Defect 4a (cross-process watcher dedup) — PR #13

Still open, unfixed by anything merged so far:
- ❌ Defect 2 — per-directory FD amplification (1,660 FDs for 315 files)
- ❌ Defect 3 — restart path leaks previous generations
- ❌ Defect 4b — disabled-guard bypass in auto_start_watch_opportunistic (confirmed still skip_disabled_check=true hardcoded)
- ❌ Defect 5 — --ui workers never exit after client disconnect

johnintuit:
@murari316 can we merge this ?

johnintuit:
@pradeepmouli pls add test case

murari316 (approved):
LGTM. Root cause correct: Cargo feature unification leaks macos_kqueue to all crates. FSEvents is right default (O(1) FDs). Canonicalize fix in shared function protects all callers. Regression test solid. Ship it.

murari316 (approved):
LGTM. Root cause correct: Cargo feature unification leaks macos_kqueue to all crates. FSEvents is right default. Canonicalize fix in shared function protects all callers. Regression test solid.

PR #12, opened 2026-07-17
fix: pass scip-enrich language list as a positional argument, not a flag

### LINKED ISSUE #11: SCIP enrichment never runs: spawn_scip_child_process passes --languages as a flag but the subcommand expects a positional argument
### Description

Background SCIP enrichment (the step that runs `scip-typescript`, `scip-python`, `rust-analyzer`, `scip-java`, `scip-go`, `scip-ruby`, `scip-clang`, `scip-php` after `index`) never actually runs for any language, on any project, silently. Every `index` prints something like:

```
SCIP enrichment starting in background (8 indexer(s): scip-typescript, scip-python, rust-analyzer, scip-java, scip-go, scip-ruby, scip-clang, scip-php)...
```

but the spawned child process exits immediately with a CLI argument-parsing error, before it reaches any per-language indexer logic. The failure is invisible unless you go read `.infigraph/scip-enrich.log`.

### Evidence

`.infigraph/scip-enrich.log` after running `index` on a real project:

```
error: unexpected argument '--languages' found

  tip: to pass '--languages' as a value, use '-- --languages'

Usage: infigraph scip-enrich <LANGUAGES>

For more information, try '--help'.
```

Verified directly with a debug build of the CLI:

```
$ infigraph scip-enrich --languages "typescript,python"      # what the code actually sends
error: unexpected argument '--languages' found
...
exit code: 2

$ infigraph scip-enrich "typescript,python"                  # positional, as the subcommand expects
Auto-SCIP: installing scip-typescript via npm...
Auto-SCIP: installing scip-python via npm...
```

The corrected (positional) invocation immediately proceeds into real per-language indexer dispatch — so `scip_download::indexers_for_languages` and the rest of `cmd_scip_enrich` work correctly once given valid arguments. The bug is isolated entirely to how the argument is passed at the spawn call site.

### Root cause

`spawn_scip_child_process` in `crates/infigraph-cli/src/index.rs:249-252` shells out to the current executable as:

```rust
let _ = std::process::Command::new(exe)
    .arg("scip-enrich")
    .arg("--languages")
    .arg(&langs)
    ...
```

But the `ScipEnrich` subcommand is declared in `crates/infigraph-cli/src/main.rs:582-585` with `languages` as a **positional** clap argument, not a flag:

```rust
ScipEnrich {
    /// Comma-separated detected languages
    languages: String,
},
```

There is no `#[arg(long)]` (or similar) annotation, so clap never registers a `--languages` flag at all — passing `--languages` as a literal argument token is rejected outright with the usage error above. Every spawned child dies on argument parsing before `cmd_scip_enrich` (`crates/infigraph-cli/src/index.rs:591`) is ever reached.

Because the child is spawned detached (`stdout(Stdio::null())`, stderr redirected only to `.infigraph/scip-enrich.log`) and the parent (`index`) returns immediately regardless of the child's outcome, this has presumably been silently broken for some time — SCIP-based enrichment (which the docs describe as part of the normal indexing pipeline) has not actually executed via this path at all.

### Suggested fix

One-line change at the call site — drop the `--languages` flag arg, passing the value positionally:

```rust
let _ = std::process::Command::new(exe)
    .arg("scip-enrich")
    .arg(&langs)
    ...
```

Given the silent-failure blast radius, it's also worth considering:
- Checking the spawned child's exit status (or at minimum logging a warning) instead of using `let _ =` on the `Command::spawn()`/ignoring the child entirely.
- A regression test asserting the exact argv passed to the respawned CLI matches what `Commands::ScipEnrich`'s clap definition actually accepts (would have caught this immediately).

pradeepmouli:
## Summary

Fixes #11. Background SCIP enrichment (`scip-typescript`, `scip-python`, `rust-analyzer`, `scip-java`, `scip-go`, `scip-ruby`, `scip-clang`, `scip-php` — whatever `index` prints as "starting in background") has never actually run, on any project, for any language.

`spawn_scip_child_process` (`crates/infigraph-cli/src/index.rs`) respawns the CLI as a detached child:

```rust
Command::new(exe).arg("scip-enrich").arg("--languages").arg(&langs)
```

but `Commands::ScipEnrich` declares `languages` as a **positional** argument (no `#[arg(long)]`), so clap never registered a `--languages` flag. Every respawned child dies immediately:

```
error: unexpected argument '--languages' found
Usage: infigraph scip-enrich <LANGUAGES>
```

Verified directly with a debug build:

```
$ infigraph scip-enrich --languages "typescript,python"   # what the code sends today
error: unexpected argument '--languages' found
exit code: 2

$ infigraph scip-enrich "typescript,python"                # positional, as declared
Auto-SCIP: installing scip-typescript via npm...
Auto-SCIP: installing scip-python via npm...
```

So the per-language dispatch logic in `cmd_scip_enrich`/`scip_download::indexers_for_languages` is correct — this is purely a wrong-argument-style bug at the spawn call site. Because the child is spawned fully detached (stdout/stdin null, stderr only to `.infigraph/scip-enrich.log`) and the parent returns immediately regardless of outcome, the failure has been completely silent — the "SCIP enrichment starting in background (N indexers: ...)" message prints unconditionally *before* the spawn.

This PR extracts the args into `scip_enrich_args()` and fixes it to pass the language list positionally.

## Test plan

- [x] `cargo test -p infigraph-cli` passes (53 tests, single-threaded and parallel)
- [x] `cargo clippy -p infigraph-cli --all-targets -- -D warnings` passes with zero new warnings (allowing only two lints that reproduce identically on unpatched `main` with this toolchain — a `useless_borrows_in_formatting` in `vuln/mod.rs` and a `type_complexity` warning in `mcp/tools/search.rs`, both in files untouched by this change)
- [x] `cargo fmt --all -- --check` introduces zero new diffs vs. `main`
- [x] Tested manually:
  - Added `scip_enrich_args_parse_as_positional_language`, which parses the exact args `spawn_scip_child_process` sends through the real `Cli`/`Commands` clap definition (no process spawn needed) and asserts they resolve to `Commands::ScipEnrich { languages }`. Confirmed it fails with the old `--languages` flag (same `UnknownArgument` error as the real bug) and passes with the fix.
  - Ran both the broken and corrected invocations directly against a debug build of the CLI binary (output above) to confirm the actual runtime behavior, not just the unit test.

## Notes

- Independent of #9/#10 (the macOS FD-leak fix) — different subsystem, no overlap in files touched.
- Given the silent-failure blast radius here, it might be worth a follow-up to have `spawn_scip_child_process` at least log a warning on a non-zero/error child exit rather than `let _ = ...spawn()`, so a future regression like this surfaces louder than a log file nobody reads. Left out of this PR to keep the diff minimal and focused on the actual bug; happy to add if useful.

### DIFF
--- a/crates/infigraph-cli/src/index.rs
+++ b/crates/infigraph-cli/src/index.rs
@@ -246,13 +246,32 @@ fn spawn_scip_child_process(root: &Path, detected_languages: &std::collections::
         Err(_) => std::process::Stdio::null(),
     };
 
-    let _ = std::process::Command::new(exe)
+    match std::process::Command::new(exe)
         .args(scip_enrich_args(&langs))
         .current_dir(root)
         .stdin(std::process::Stdio::null())
         .stdout(std::process::Stdio::null())
         .stderr(stderr_target)
-        .spawn();
+        .spawn()
+    {
+        Ok(mut child) => {
+            // spawn() only reports failure to launch (missing binary, exec
+            // permission). It says nothing about the child crashing or
+            // exiting nonzero afterward — exactly the failure shape of the
+            // bug this function used to hit silently (the child launched
+            // fine and died instantly inside clap's parser). Wait on it from
+            // a detached thread so this function still returns immediately,
+            // but any future silent-death cause surfaces a warning instead
+            // of only leaving a trace in a log nobody's prompted to open.
+            let log_path = log_path.clone();
+            std::thread::spawn(move || {
+                if let Some(msg) = scip_enrich_exit_message(child.wait(), &log_path) {
+                    eprintln!("{msg}");
+                }
+            });
+        }
+        Err(e) => eprintln!("  Warning: failed to spawn scip-enrich: {e}"),
+    }
 
     eprintln!("  Log: {}", log_path.display());
 }
@@ -263,6 +282,25 @@ fn spawn_scip_child_process(root: &Path, detected_languages: &std::collections::
 /// without spawning a process.
 fn scip_enrich_args(langs: &str) -> Vec<String> {
     vec!["scip-enrich".to_string(), langs.to_string()]
+}
+
+/// Decides what (if anything) to warn about after waiting on the detached
+/// scip-enrich child. Extracted from the wait thread so it's testable
+/// without spawning a real process — `current_exe()` in `spawn_scip_child_process`
+/// resolves to the test binary itself under `cargo test`, not `infigraph`,
+/// so the full spawn path can't be exercised end-to-end in a unit test.
+fn scip_enrich_exit_message(
+    status: std::io::Result<std::process::ExitStatus>,
+    log_path: &Path,
+) -> Option<String> {
+    match status {
+        Ok(status) if !status.success() => Some(format!(
+            "warning: scip-enrich exited with {status} — see {}",
+            log_path.display()
+        )),
+        Err(e) => Some(format!("warning: failed to wait on scip-enrich: {e}")),
+        _ => None,
+    }
 }
 
 pub(crate) const CI_ENV_VARS: &[&str] = &[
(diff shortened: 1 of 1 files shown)


crates/infigraph-cli/src/index.rs
johnintuit: The core fix is correct and good for issue - 11. 

One follow-up worth doing in this same PR or a fast-follow: spawn_scip_child_process still discards the **spawn**() result:

let _ = std::process::Command::new(exe)
    .args(scip_enrich_args(&langs))
    .current_dir(root)
    .stdin(std::process::Stdio::null())
    .stdout(std::process::Stdio::null())
    .stderr(stderr_target)
    .**spawn**();

spawn()'s Result only reports failure to launch the process (binary missing, exec permission denied) — it says nothing about the child crashing or exiting nonzero afterward. That's exactly the failure shape of the bug this PR fixes: the child launched fine and died instantly inside clap's parser, and spawn() alone would never have caught that. Discarding the result means any other silent-child-death cause (missing scip-* indexer binary, permission error, panic) will reproduce the same invisible-failure symptom, with the same "SCIP enrichment starting in background" message printing regardless.

Suggested fix — wait on the child from a detached thread so this function still returns immediately, and warn on spawn failure or nonzero exit:

match std::process::Command::new(exe)
    .args(scip_enrich_args(&langs))
    .current_dir(root)
    .stdin(std::process::Stdio::null())
    .stdout(std::process::Stdio::null())
    .stderr(stderr_target)
    .spawn()
{
    Ok(mut child) => {
        let log_path = log_path.clone();
        std::thread::spawn(move || match child.wait() {
            Ok(status) if !status.success() => {
                eprintln!(
                    "warning: scip-enrich exited with {status} — see {}",
                    log_path.display()
                );
            }
            Err(e) => eprintln!("warning: failed to wait on scip-enrich: {e}"),
            _ => {}
        });
    }
    Err(e) => eprintln!("  Warning: failed to spawn scip-enrich: {e}"),
}

This keeps the "detached, non-blocking" behavior intact — the parent returns immediately either way — but now any future silent-death bug in this path surfaces a warning instead of only leaving a trace in .infigraph/scip-enrich.log that nobody's prompted to open. Not a blocker for merging the current fix, just flagging so it doesn't get lost.

johnintuit (changes requested):

johnintuit:
@pradeepmouli pls add test case also

pradeepmouli:
Thanks for the review — pushed a follow-up commit implementing exactly what you suggested: `spawn_scip_child_process` now waits on the child from a detached thread and warns on a nonzero exit or spawn failure, instead of discarding the result. Added `scip_enrich_exit_message()` as the testable decision logic (extracted since `current_exe()` resolves to the *test* binary under `cargo test`, not `infigraph`, so the full spawn→wait path can't be exercised end-to-end in a unit test — this function lets the wait/warn decision be tested directly by constructing `ExitStatus` values, without spawning a real process).

Full local test run (fresh clone, this branch, no CI available on this fork so pasting the actual output rather than just describing it):

```
$ cargo test -p infigraph-cli --bin infigraph
running 56 tests
...
test index::tests::scip_enrich_args_parse_as_positional_language ... ok
test index::tests::scip_enrich_exit_message_silent_on_success ... ok
test index::tests::scip_enrich_exit_message_warns_on_nonzero_exit ... ok
test index::tests::scip_enrich_exit_message_warns_on_wait_error ... ok
...
test result: ok. 56 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.08s
```

(53 pre-existing + 3 new: the two from your suggested follow-up, plus a wait()-error case.)

Confirmed these are real regression guards, not decorative — temporarily disabled the nonzero-exit branch in `scip_enrich_exit_message` and reran:

```
thread 'index::tests::scip_enrich_exit_message_warns_on_nonzero_exit' panicked at crates/infigraph-cli/src/index.rs:960:9:
expected a warning mentioning the exit status and log path, got None
test index::tests::scip_enrich_exit_message_warns_on_nonzero_exit ... FAILED
test result: FAILED. 3 passed; 1 failed; 0 ignored; 0 measured; 52 filtered out; finished in 0.00s
```

Restored, reran, all green again.

Also confirmed zero new `cargo fmt --all -- --check` diffs and a clean `cargo clippy -p infigraph-cli --all-targets -- -D warnings` (allowing only the two pre-existing lints that reproduce identically on unpatched `main` with this toolchain — a `useless_borrows_in_formatting` in `vuln/mod.rs` and a `type_complexity` warning in `mcp/tools/search.rs`, both in files untouched by this PR).

johnintuit (approved):

PR #13, opened 2026-07-17
fix: prevent duplicate MCP watchers on the same project across processes

pradeepmouli:
## Summary

Follow-up from the investigation in #9/#10 (macOS FD leak) and #11/#12 (scip-enrich arg fix) — a related but independent bug in MCP watcher lifecycle, not fixed by either of those.

`tool_watch_project` never acquired `.infigraph/watch.lock` itself. The only existing guard, `cli_watcher_holds_lock` (used solely by `auto_start_watch`'s pre-check), just *peeked* at whether a CLI-started watcher held the lock — it never gated one MCP worker's watcher against another MCP worker's, since neither ever took the lock. Every worker process that called `index_project`, `group_index`, or `search` therefore built its own duplicate watcher set for the same project — each independently walking the tree and reindexing on every file change.

This PR makes `tool_watch_project` acquire that same per-project lock itself before spawning its watcher thread, and hold it for the watcher's lifetime — mirroring exactly what the CLI's own `cmd_watch`/`ensure_watcher_running` already do via `acquire_watch_lock` (`info_commands.rs:412-425`). A second attempt to watch an already-watched project (from another MCP worker, or another explicit `watch_project` call) now fails with an "already running" error, which flows into `auto_start_watch`'s existing `Err(e) => None` branch — so the documented "returns `None` when another watcher already covers this project" contract is unchanged for existing callers/tests, just now correctly enforced against MCP-vs-MCP duplication too, not only MCP-vs-CLI.

Went with graceful decline (fail to acquire → don't start a duplicate) rather than kill-and-take-over: `try_lock_exclusive` is an OS-level advisory flock, which the kernel releases automatically the instant a holding process dies — there's no "stale lock from a dead process" scenario to justify forcibly breaking a live holder's lock, and doing so would risk killing a watcher that's still doing real work for a different, legitimate client.

Acquisition retries briefly (up to ~500ms) since a watcher that was just told to stop may take up to its ~200ms poll interval to actually exit and release the lock; without this, a legitimate restart attempt landing in that handover window would spuriously fail.

### Test infra fix included

While verifying this, found a latent race in the test suite itself: `stop_all_watchers()` (duplicated in `watcher_concurrency.rs` and `watcher_reindex.rs`) only *signals* watcher threads to stop, without waiting for them to actually exit. Once `tool_watch_project` started holding a real lock, this raced legitimately — a test that stopped a watcher and immediately re-watched the same path could hit the handover window and get a false "already running" failure (reproduced this exact failure with `test_mcp_starts_when_no_cli_lock` before fixing it). Both copies of the helper now wait, with a bounded timeout, for the stopped watcher's lock to be confirmed released before returning.

## Test plan

- [x] `cargo test -p infigraph-mcp --test watcher_concurrency --test watcher_reindex` passes (25 tests, single-threaded)
- [x] `cargo clippy -p infigraph-mcp --all-targets -- -D warnings` passes with zero new warnings (allowing only the two pre-existing lints confirmed to reproduce identically on unpatched `main`, in files untouched by this change)
- [x] `cargo fmt --all -- --check` introduces zero new diffs vs. `main`
- [x] Tested manually:
  - Added `test_second_watch_project_call_declines_when_already_watching`, which calls `tool_watch_project` twice directly for the same path — bypassing `auto_start_watch`'s own in-process `is_watching` map, so the test proves the *lock* itself (not in-process state a second real worker process wouldn't share) is what prevents the duplicate. Confirmed it fails when the lock acquisition is temporarily bypassed and passes with the fix.
  - Confirmed the pre-existing `test_mcp_skips_when_cli_lock_held` / `test_mcp_starts_when_no_cli_lock` / `test_no_stale_warning_with_cli_watcher` tests all pass unchanged — the `None`-on-decline contract for the CLI-lock case is preserved exactly.

## Notes

- Independent of #10/#12 — different files, no overlap, can merge in any order.
- `cli_watcher_holds_lock` is removed (dead code once its only call site — the now-redundant pre-check in `auto_start_watch_inner` — is gone; `tool_watch_project`'s own acquire-attempt subsumes it for both CLI-held and MCP-held locks).

### DIFF
--- a/crates/infigraph-mcp/src/tools/watch.rs
+++ b/crates/infigraph-mcp/src/tools/watch.rs
@@ -83,10 +83,6 @@ fn auto_start_watch_inner(path: &str, skip_disabled_check: bool) -> Option<Strin
         return None;
     }
 
-    if cli_watcher_holds_lock(&root) {
-        return None;
-    }
-
     let args = serde_json::json!({
         "path": path,
         "auto_resolve": true,
@@ -104,25 +100,35 @@ fn auto_start_watch_inner(path: &str, skip_disabled_check: bool) -> Option<Strin
     }
 }
 
-fn cli_watcher_holds_lock(root: &std::path::Path) -> bool {
+/// Acquire the per-project watch lock (`.infigraph/watch.lock`), the same
+/// lock the CLI's `infigraph watch` holds for its lifetime. Retries briefly:
+/// a watcher that was just told to stop may take up to its poll interval
+/// (~200ms) to actually exit and release this lock, and without a retry
+/// window a start attempt landing in that gap would spuriously fail.
+fn acquire_project_watch_lock(lock_path: &std::path::Path) -> Result<std::fs::File> {
     use fs2::FileExt;
-    let lock_path = root.join(".infigraph").join("watch.lock");
-    let file = match std::fs::OpenOptions::new()
-        .create(false)
+    if let Some(parent) = lock_path.parent() {
+        std::fs::create_dir_all(parent)?;
+    }
+    let file = std::fs::OpenOptions::new()
+        .create(true)
         .write(true)
         .truncate(false)
-        .open(&lock_path)
-    {
-        Ok(f) => f,
-        Err(_) => return false,
-    };
-    match file.try_lock_exclusive() {
-        Ok(()) => {
-            let _ = file.unlock();
-            false
+        .open(lock_path)?;
+
+    const ATTEMPTS: u32 = 10;
+    const RETRY_DELAY: std::time::Duration = std::time::Duration::from_millis(50);
+    let mut last_err = None;
+    for _ in 0..ATTEMPTS {
+        match file.try_lock_exclusive() {
+            Ok(()) => return Ok(file),
+            Err(e) => {
+                last_err = Some(e);
+                std::thread::sleep(RETRY_DELAY);
+            }
         }
-        Err(_) => true,
     }
+    Err(last_err.unwrap().into())
 }
 
 pub fn tool_watch_project(args: &Value) -> Result<String> {
@@ -145,6 +151,15 @@ pub fn tool_watch_project(args: &Value) -> Result<String> {
         .canonicalize()
         .context("invalid path")?;
     let root_str = root.to_string_lossy().replace('\\', "/");
+
+    let lock_path = root.join(".infigraph").join("watch.lock");
+    let watch_lock = acquire_project_watch_lock(&lock_path).map_err(|_| {
+        anyhow::anyhow!(
+            "another watcher is already running for {root_str} \
+             (CLI `infigraph watch` or another MCP worker) — \
+             use get_watch_status to check, or stop it first"
+        )
+    })?;
 
     let (stop_tx, stop_rx) = mpsc::channel::<()>();
     let watcher_id = format!(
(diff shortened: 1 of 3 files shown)


johnintuit (approved):
Looks good, we can merge, don't close reported issues as it is not fully fixed

johnintuit:
@pradeepmouli pls add test case also

pradeepmouli:
No CI runs on this fork, so posting the actual local test output as evidence rather than just describing it — fresh clone, this exact branch, no changes since the last push.

```
$ cargo test -p infigraph-mcp --test watcher_concurrency --test watcher_reindex -- --test-threads=1

running 8 tests
test test_graph_tools_with_active_watchers ... ok
test test_graph_tools_with_group_watchers ... ok
test test_mcp_skips_when_cli_lock_held ... ok
test test_mcp_starts_when_no_cli_lock ... ok
test test_no_stale_warning_with_cli_watcher ... ok
test test_no_stale_warning_with_mcp_watcher ... ok
test test_search_auto_starts_watcher_when_none_running ... ok
test test_second_watch_project_call_declines_when_already_watching ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 115.19s

running 17 tests
test test_auto_start_doc_watch_no_duplicates ... ok
test test_auto_start_watch_no_duplicates ... ok
test test_code_watcher_branch_switch_existing_dirs ... ok
test test_code_watcher_branch_switch_new_dirs ... ok
test test_code_watcher_cross_file_auto_resolve ... ok
test test_code_watcher_grammar_plugin_extensions ... ok
test test_code_watcher_handles_dir_removal ... ok
test test_code_watcher_ignores_excluded_dirs ... ok
test test_code_watcher_reindexes_modified_file ... ok
test test_code_watcher_reindexes_new_file_existing_dir ... ok
test test_code_watcher_reindexes_new_file_new_dir ... ok
test test_code_watcher_sentinel_stop ... ok
test test_doc_watcher_prunes_stale_docs ... ok
test test_doc_watcher_reindexes_new_doc ... ok
test test_doc_watcher_reindexes_new_doc_new_dir ... ok
test test_doc_watcher_reindexes_no_concurrent_read ... ok
test test_is_watching_lifecycle ... ok

test result: ok. 17 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 149.89s
```

25/25 pass, including the 3 pre-existing tests that specifically cover the CLI-held-lock contract (`test_mcp_skips_when_cli_lock_held`, `test_mcp_starts_when_no_cli_lock`, `test_no_stale_warning_with_cli_watcher`) — confirming the `None`-on-decline behavior for that case is unchanged — plus the new `test_second_watch_project_call_declines_when_already_watching`, which is the one that actually exercises this PR's fix: it calls `tool_watch_project` twice directly for the same path, bypassing `auto_start_watch`'s own in-process `is_watching` map, so it can only pass because of the lock acquired inside `tool_watch_project` itself — not in-process state a second real MCP worker process wouldn't share.

To confirm that test is load-bearing rather than decorative: I temporarily bypassed the lock acquisition in `acquire_project_watch_lock` (made it always succeed regardless of contention) and reran just that test:

```
thread 'test_second_watch_project_call_declines_when_already_watching' panicked at crates/infigraph-mcp/tests/watcher_concurrency.rs:476:5:
a second watch_project call for an already-watched path must decline, not start a duplicate watcher
test test_second_watch_project_call_declines_when_already_watching ... FAILED
```

Restored the real lock acquisition, reran, back to green.

Also confirmed `cargo fmt --all -- --check` introduces zero new diffs vs. `main` on this branch (matches the same 76-file pre-existing baseline as the other PRs — this toolchain's rustfmt disagrees with whatever version last ran in CI, unrelated to this change), and `cargo clippy -p infigraph-mcp --all-targets -- -D warnings` is clean (allowing only the two pre-existing lints that reproduce identically on unpatched `main`).

PR #20, opened 2026-07-19
fix: don't wipe the graph on lock contention from another live process

pradeepmouli:
## Summary

`Infigraph::init()` treats every `KuzuBackend::open()` failure as
corruption and wipes + rebuilds the database -- including Kuzu's own
`"Could not set lock on file"` error, which is plain lock contention from
another live process (e.g. a running `infigraph watch` that already has
the database open), not corruption at all.

Concretely: run `infigraph watch` in one terminal, then run any other
`infigraph` command (`stats`, `index`, ...) against the same project in
another terminal while the watcher is alive. The second command's open
attempt fails with a lock error, `init()` treats that as corruption, wipes
`.infigraph/graph`, and rebuilds from whatever partial state is visible at
that moment -- destroying the watcher's live data out from under it. This
happened twice in a row against a real project in my own use.

## Fix

Add `is_lock_contention_error()`, which checks for Kuzu's lock-contention
error text specifically (`GraphStore::open` collapses the underlying Kuzu
error into a stringified `anyhow::Error` before it reaches `Infigraph::init`,
so there's no structured error variant to match on -- only Kuzu's own error
text). On lock contention, return a clear
`"graph is locked by another infigraph process ... not corrupted, so it was
left untouched"` error instead of wiping. Genuine corruption (e.g. the
`"Database ID ... does not match"` error from a stale/mismatched WAL
checkpoint file) still triggers the existing wipe-and-rebuild recovery --
only lock contention is exempted.

## Test plan

- [x] Unit tests: `is_lock_contention_error_matches_kuzu_lock_message`,
      `is_lock_contention_error_does_not_match_genuine_corruption`
- [x] Full `infigraph-core` unit test suite (268 tests) passes, no regressions
- [x] End-to-end with two real, separate OS processes: started a real
      `infigraph watch` holding a graph open, ran `infigraph stats` against
      the same project in a second process. Before this fix: the second
      call's failure wiped the watcher's graph. After: it returns the clear
      error above and the watcher's data survives intact (confirmed via
      `stats` on the still-running watcher after the second call).

### DIFF
--- a/crates/infigraph-core/src/lib.rs
+++ b/crates/infigraph-core/src/lib.rs
@@ -55,6 +55,18 @@ pub(crate) fn escape_str(s: &str) -> String {
 pub fn build_hash() -> &'static str {
     env!("INFIGRAPH_BUILD_HASH")
 }
+
+/// Kuzu's IO-layer error for "another process holds this database's lock"
+/// (see docs.ladybugdb.com/concurrency) is lock contention, not corruption.
+/// `GraphStore::open` collapses the underlying Kuzu error into a stringified
+/// `anyhow::Error` before it reaches `Infigraph::init`, so there's no
+/// structured error variant to match on here -- only Kuzu's own error text.
+/// This was previously indistinguishable from genuine corruption, so a
+/// second `infigraph` process opening a graph while a watcher already had
+/// it open would trigger `wipe_graph`, destroying the watcher's live data.
+fn is_lock_contention_error(err: &anyhow::Error) -> bool {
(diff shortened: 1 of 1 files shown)


johnintuit (approved):
Looks good

johnintuit:
@murari316 pls merge this

johnintuit:
no, there is conflicting files now with main

johnintuit:
@pradeepmouli can you pls update conflicitng file?

pradeepmouli:
Should be good

PR #21, opened 2026-07-19
fix: extract INHERITS edges for TypeScript and Rust

pradeepmouli:
## Summary

Both languages silently produced zero inheritance edges:

- `typescript/relations.scm` had no `@inherit.child`/`@inherit.parent`
  capture at all -- only calls and imports -- unlike `python/relations.scm`
  and `javascript/relations.scm`, which both have working patterns for
  this. Every `class X extends Y`, `interface X extends Y`, and `class X
  implements Y` was invisible to the graph.
- `rust/relations.scm` had a comment describing intent to capture `impl
  Trait for Type` as a relationship ("We capture impl blocks as a
  relationship"), but no actual query pattern was ever written. Every
  trait impl in every Rust codebase produced zero `INHERITS` edges.

Confirmed against this repo's own code: `impl GraphBackend for
KuzuBackend` (`crates/infigraph-core/src/graph/kuzu_backend.rs`) had no
corresponding graph edge before this fix. Also confirmed against a real
external TypeScript repo's `InputProps extends
React.ComponentProps<'input'>`.

This also directly affects the pattern-detection module
(`patterns/mod.rs`) -- Strategy and Decorator both rely on `INHERITS`
edges as their core structural signal, and were effectively blind for
TypeScript and Rust codebases until this fix.

## Fix

Verified the tree-sitter grammar node structure against the vendored
`tree-sitter-typescript`/`tree-sitter-rust` `node-types.json` rather than
guessing:
- TypeScript: `class_declaration` -> `class_heritage` -> `extends_clause`
  (value field) / `implements_clause`; `interface_declaration` ->
  `extends_type_clause` (type field).
- Rust: `impl_item` -> `trait`/`type` fields.

## Test plan

- [x] Compiles clean, no language-pack load warnings (confirms query
      syntax doesn't break existing extraction for either language)
- [x] Full `infigraph-core` (268) and `infigraph-languages` (11) test
      suites pass
- [x] Two new regression tests added to
      `infigraph-languages/tests/registry_integration.rs`:
      `test_extraction_rust_impl_trait_produces_inherits_edge`,
      `test_extraction_typescript_inheritance_produces_edges` (covers
      class extends, interface extends, and class implements)
- [x] End-to-end against a synthetic project covering all four patterns
      (class extends, interface extends, class implements, Rust trait
      impl) -- all four produce correct `INHERITS` edges
- [x] Applied to 7 real repos and reindexed: `Inherits` counts went from
      0 (or near-0) to real, non-trivial counts across the board (one repo
      stayed at 0, correctly -- its only inheritance target is the
      external/built-in `Error` class, which has no `Symbol` node in the
      graph to resolve against, same as unresolved external `Calls` edges)

### DIFF
--- a/crates/infigraph-languages/languages/typescript/relations.scm
+++ b/crates/infigraph-languages/languages/typescript/relations.scm
@@ -14,22 +14,23 @@
 (import_statement
   source: (string) @import.module)
 
-; Class inheritance: class Foo extends Bar
+; Class inheritance: class Foo extends Bar (Bar may be identifier or member_expression)
 (class_declaration
   name: (type_identifier) @inherit.child
   (class_heritage
     (extends_clause
-      value: (identifier) @inherit.parent)))
+      value: (_) @inherit.parent)))
 
-; Interface inheritance: interface Foo extends Bar
+; Interface inheritance: interface Foo extends Bar (Bar may be type_identifier,
+; generic_type, or nested_type_identifier)
 (interface_declaration
   name: (type_identifier) @inherit.child
   (extends_type_clause
-    type: (type_identifier) @inherit.parent))
+    type: (_) @inherit.parent))
 
 ; Class implements: class Foo implements Bar
 (class_declaration
   name: (type_identifier) @inherit.child
   (class_heritage
     (implements_clause
-      (type_identifier) @inherit.parent)))
+      (_) @inherit.parent)))
--- a/crates/infigraph-languages/languages/typescript/relations.scm
+++ b/crates/infigraph-languages/languages/typescript/relations.scm
@@ -13,3 +13,23 @@
 ; Import statements
 (import_statement
   source: (string) @import.module)
+
+; Class inheritance: class Foo extends Bar
+(class_declaration
+  name: (type_identifier) @inherit.child
+  (class_heritage
+    (extends_clause
+      value: (identifier) @inherit.parent)))
+
+; Interface inheritance: interface Foo extends Bar
+(interface_declaration
+  name: (type_identifier) @inherit.child
+  (extends_type_clause
+    type: (type_identifier) @inherit.parent))
+
+; Class implements: class Foo implements Bar
+(class_declaration
+  name: (type_identifier) @inherit.child
+  (class_heritage
+    (implements_clause
+      (type_identifier) @inherit.parent)))
--- a/crates/infigraph-languages/tests/registry_integration.rs
+++ b/crates/infigraph-languages/tests/registry_integration.rs
@@ -142,6 +142,31 @@ fn test_extraction_smoke_rust() {
     assert!(names.contains(&"main"), "should extract main: {names:?}");
 }
 
+/// Regression test: rust/relations.scm had a comment describing intent to
+/// capture `impl Trait for Type` as an INHERITS relationship, but no actual
+/// query pattern was ever written -- every trait impl in every Rust codebase
+/// silently produced zero INHERITS edges (confirmed against infigraph's own
+/// `impl GraphBackend for KuzuBackend`, which had no corresponding edge).
+#[test]
+fn test_extraction_rust_impl_trait_produces_inherits_edge() {
+    use infigraph_core::model::RelationKind;
+
+    let registry = bundled_registry().unwrap();
+    let pack = registry.for_extension(".rs").unwrap();
+
+    let source = b"trait Greet {\n    fn hello(&self);\n}\nstruct Person;\nimpl Greet for Person {\n    fn hello(&self) {}\n}\n";
+    let extraction = infigraph_core::extract::extract_file("test.rs", source, pack)
+        .expect("extraction should succeed");
+
+    assert!(
+        extraction.relations.iter().any(|r| r.kind == RelationKind::Inherits
+            && r.source_id.contains("Person")
+            && r.target_id.contains("Greet")),
+        "expected an INHERITS edge from Person to Greet, got: {:?}",
+        extraction.relations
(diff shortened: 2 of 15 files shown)


crates/infigraph-languages/languages/typescript/relations.scm
```diff
@@ -13,3 +13,23 @@
 ; Import statements
 (import_statement
   source: (string) @import.module)
+
+; Class inheritance: class Foo extends Bar
+(class_declaration
+  name: (type_identifier) @inherit.child
+  (class_heritage
+    (extends_clause
```
johnintuit, line 21: Idea is to draw an arrow ("INHERITS edge") in the code graph whenever one TypeScript class/interface extends or implements another. It works for the simple case (class Dog extends Animal) but breaks for two common real-world patterns because the tree-sitter query is too narrow — it only accepts a bare name, not the fuller syntax tree-sitter actually produces for more complex code.

**Bug 1 — extends Something.Else (dotted/member names) gets dropped.**
Real code: class MyComponent extends React.Component — extremely common in any React class-component codebase. The query says "parent must be a plain identifier node." But when you write React.Component, tree-sitter doesn't parse that as one identifier — it parses it as a member_expression (object=React, property=Component). The query's identifier-only filter doesn't match a member_expression, so the whole pattern silently fails and no edge gets created. Same bug class the PR is fixing, just missed a case.

**Bug 2 — implements Something<T> (generics) gets dropped.**
Real code: class Square implements Drawable<Size>. Query says the implemented name must be a plain type_identifier node. But Drawable<Size> parses as a generic_type node (which wraps a type_identifier inside it), not a bare type_identifier. Query doesn't match generic_type, so again — no edge, silently.

- Line 22: value: (identifier) @inherit.parent — bug 1, needs to accept member-expressions too.
- Line 34: (type_identifier) @inherit.parent)) — bug 2, needs to accept generic_type/nested_type_identifier too.

crates/infigraph-languages/languages/typescript/relations.scm
```diff
@@ -13,3 +13,23 @@
...
+      value: (identifier) @inherit.parent)))
+
+; Interface inheritance: interface Foo extends Bar
+(interface_declaration
+  name: (type_identifier) @inherit.child
+  (extends_type_clause
+    type: (type_identifier) @inherit.parent))
```
johnintuit: L28 (typescript/relations.scm, interface extends): 🔴 bug: `type: (type_identifier)` misses generic_type and nested_type_identifier — grammar's extends_type_clause.type field explicitly allows both. `interface Circle extends Shape<T>` or `interface Foo extends ns.Bar` silently drops the edge. Fix: `type: (_) @inherit.parent`. Add test for `extends Shape<T>`.

rust/relations.scm (impl_item query): 🔴 bug: both `trait: (type_identifier)` and `type: (type_identifier)` are too narrow — grammar allows trait field to be generic_type/scoped_type_identifier/type_identifier, and type field is the broad `_type` supertype. `impl std::fmt::Display for MyType` or `impl Iterator<Item=T> for MyType` (both common Rust patterns) silently drop the edge. Fix: `trait: (_) @inherit.parent` and `type: (_) @inherit.child`. Add tests for scoped trait path and generic trait.

crates/infigraph-languages/tests/registry_integration.rs
```diff
@@ -142,6 +142,34 @@ fn test_extraction_smoke_rust() {
     assert!(names.contains(&"main"), "should extract main: {names:?}");
 }
 
+/// Regression test: rust/relations.scm had a comment describing intent to
+/// capture `impl Trait for Type` as an INHERITS relationship, but no actual
+/// query pattern was ever written -- every trait impl in every Rust codebase
+/// silently produced zero INHERITS edges (confirmed against infigraph's own
+/// `impl GraphBackend for KuzuBackend`, which had no corresponding edge).
+#[test]
```
johnintuit, line 150: Test recommendation — don't just add 4 more one-off assertions. Root cause is "query hardcodes narrowest node type instead of matching the field's actual declared type set." I suggest to 

1. Fix all 4 queries by loosening to (_) (wildcard, still respects the field name, just not the concrete node type) — this is the actual fix, not a workaround.
2. Add table-driven test cases per language covering: plain identifier, member/scoped path, generic — 3 cases × (TS extends, TS implements, TS interface-extends, Rust impl-trait) = up to 12 cases, but can dedupe into one parametrized test per language rather than 12 separate #[test] fns.
3. Not required to enumerate every grammar alternative by hand — using (_) sidesteps needing to keep the query in sync with grammar node-type additions later.

johnintuit (changes requested):

murari316 (approved):
Reviewed diff in full — TypeScript/Rust INHERITS extraction verified against actual tree-sitter node-types.json structure, not guessed. Regression tests cover class extends, interface extends, class implements, and Rust trait impl. Approving.

murari316:
@pradeepmouli are @johnintuit changes are addressed?

crates/infigraph-languages/tests/registry_integration.rs
pradeepmouli: Right. Perhaps it's best to match the higher-level node (e.g. class_heritage or extends_type_clause in the case of typescript, trait_item/impl_item in the case of rust) and descend inwards - instead of leaving it as a full wildcard and having to then parse that independently?

pradeepmouli: Ah - I see that's what you recommended in your comments on the other PR.

PR #22, opened 2026-07-19
fix: extract INHERITS edges for Go, Kotlin, Swift, Dart, Objective-C

pradeepmouli:
## Summary

Follow-up to #21 (TypeScript/Rust `INHERITS` extraction). Same class of
gap in five more mainstream languages -- none captured any
inheritance-like relationship at all:

- **go**: struct embedding (Go's closest analog to inheritance -- an
  anonymous field with no name, just a type). Interface satisfaction is
  implicit/structural in Go and can't be determined from syntax alone, so
  it's intentionally not covered.
- **kotlin**: class inheritance and interface implementation via
  `delegation_specifiers` (`class X : Y()` / `class X : Y`).
- **swift**: class/struct/enum inheritance and protocol conformance
  (`class X: Y`, `protocol X: Y`).
- **dart**: `extends` and `implements`.
- **objc**: class inheritance (`@interface X : Y`). Also fixes a separate,
  deeper bug in `objc/entities.scm`: `class_interface`,
  `class_implementation`, and `protocol_declaration` all used a `name:`
  field that doesn't exist on those grammar nodes (confirmed against
  `tree-sitter-objc`'s `node-types.json` -- the class name is an unlabeled
  positional child, not a field), so every Objective-C class/protocol
  produced **zero symbols at all**, not just zero inheritance edges.
  Without fixing that too, the new `relations.scm` pattern would have
  nothing to bind to.

## Fix

Verified every grammar's exact node structure against the vendored
`tree-sitter-*` `node-types.json` (and, for Objective-C, the compiled
`grammar.json` for positional/unlabeled child structure) rather than
guessing.

## Test plan

- [x] Compiles clean, no language-pack load warnings for any of the five
- [x] Full `infigraph-core` (268) and `infigraph-languages` (16) test
      suites pass
- [x] Five new regression tests in
      `infigraph-languages/tests/registry_integration.rs`, one per
      language (the Objective-C test also asserts symbols are extracted,
      covering the entities.scm fix)
- [x] End-to-end against synthetic files covering each pattern -- all
      five produce correct `INHERITS` edges (Objective-C additionally
      confirmed to now produce symbols where it previously produced zero)

### DIFF
--- a/crates/infigraph-languages/languages/go/relations.scm
+++ b/crates/infigraph-languages/languages/go/relations.scm
@@ -30,3 +30,15 @@
   (call_expression
     function: (selector_expression
       field: (field_identifier) @spawns.target))) @spawns.site
+
+; Struct embedding: type Dog struct { Animal } -- Go's closest analog to
+; inheritance. An embedded (anonymous) field has no name, only a type.
+; Interface satisfaction is implicit/structural in Go and can't be
+; determined from syntax alone, so it isn't captured here.
+(type_spec
+  name: (type_identifier) @inherit.child
+  type: (struct_type
+    (field_declaration_list
+      (field_declaration
+        !name
+        type: (type_identifier) @inherit.parent))))
--- a/crates/infigraph-languages/languages/go/relations.scm
+++ b/crates/infigraph-languages/languages/go/relations.scm
@@ -32,13 +32,15 @@
       field: (field_identifier) @spawns.target))) @spawns.site
 
 ; Struct embedding: type Dog struct { Animal } -- Go's closest analog to
-; inheritance. An embedded (anonymous) field has no name, only a type.
-; Interface satisfaction is implicit/structural in Go and can't be
-; determined from syntax alone, so it isn't captured here.
+; inheritance. An embedded (anonymous) field has no name, only a type (which
+; may be type_identifier, generic_type, or qualified_type, e.g. embedding
+; pkg.Animal or a generic Base[T]). Interface satisfaction is implicit/
+; structural in Go and can't be determined from syntax alone, so it isn't
+; captured here.
 (type_spec
   name: (type_identifier) @inherit.child
   type: (struct_type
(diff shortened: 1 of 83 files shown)


crates/infigraph-languages/languages/go/relations.scm
```diff
@@ -30,3 +30,15 @@
   (call_expression
     function: (selector_expression
       field: (field_identifier) @spawns.target))) @spawns.site
+
+; Struct embedding: type Dog struct { Animal } -- Go's closest analog to
+; inheritance. An embedded (anonymous) field has no name, only a type.
+; Interface satisfaction is implicit/structural in Go and can't be
+; determined from syntax alone, so it isn't captured here.
+(type_spec
+  name: (type_identifier) @inherit.child
```
johnintuit, line 41: Same issue like PR#21 - Kindly check here also @pradeepmouli 

crates/infigraph-languages/languages/go/relations.scm, embedded-field query (type: (type_identifier)):
🔴 bug: field_declaration.type field allows qualified_type (cross-package embed) and generic_type too, per grammar. `type Dog struct { pkg.Animal }` — very common Go pattern — parses as qualified_type, not type_identifier. Edge silently dropped. Fix: `type: (_) @inherit.parent`. Add test for package-qualified embedding.

crates/infigraph-languages/languages/dart/relations.scm, L28 and L35 ((type_identifier) inside superclass/interfaces):
🔴 bug: Dart's _type_not_void grammar rule (used by both extends/implements) wraps the type name plus optional type_arguments, and the type name itself can be library-prefixed (`pkg.Animal`). `class Dog extends Animal<T>` or `extends pkg.Animal` won't match a bare type_identifier child. Fix: `(type_identifier)` → `(_) @inherit.parent`, or match on the type node directly without descending into a specific child type. Add tests for generic and prefixed base classes.

johnintuit (changes requested):

PR #24, opened 2026-07-19
fix: retry with backoff before wiping the graph on a non-lock-contention open failure

pradeepmouli:
## Summary

**Depends on #20** — this branch is built on top of that PR's `fix/lock-contention-wipe`, since this fix extends the exact same code path (`Infigraph::init()`'s error-handling arm right after the lock-contention check). The diff currently includes #20's commit too since it isn't merged yet; it'll shrink to just this PR's own commit automatically once #20 lands.

#20 correctly avoids wiping the graph when Kuzu's open error is lock contention ("Could not set lock on file"), since that just means another live process (a watcher) already has it open. Any *other* open error — including a transient short read from a concurrent writer mid-checkpoint — fell straight through to `wipe_graph` with zero retry, on the assumption it must be genuine corruption.

This destroyed a real repo's graph tonight: a manual `scip-import` ran concurrently with a background watcher/MCP-server process touching the same graph file, producing `IO exception: Cannot read from file... 0 bytes` on the next open — not a lock message, so it wiped tens of thousands of symbols that were sitting right there on disk moments earlier.

This PR retries with backoff (200ms, 500ms, 1s) before giving up. This self-heals the transient-race case — including slower writers than a single fixed retry would tolerate, e.g. a large SCIP import still mid-checkpoint — while still correctly wiping-and-rebuilding graphs that are durably corrupt, since retries against those keep failing too. If a retry attempt comes back as lock contention specifically, it stops retrying immediately and reports that instead (now unambiguously another live process, not a checkpoint race).

## Is there a more principled fix than string-matching a retry threshold?

Investigated this directly before writing the fix:
- The underlying `lbug` (Kùzu) crate has no structured error type to match on instead of strings — open failures cross the FFI boundary as an opaque `cxx::Exception`, so there's no error code available at the Rust binding layer.
- Kùzu's own concurrency docs don't cover retry or corruption-detection at all. Their documented model is strict single-writer, with concurrent-write access meant to go through one dedicated writer process — the same pattern already used by this codebase's `Neo4jBackend`, which has no retry/wipe logic of its own since Neo4j handles concurrency server-side.

So this PR is a pragmatic safety net given that constraint, not a substitute for routing all writes through a single process (a much larger architectural change, out of scope here).

## Test plan

- [x] `cargo test -p infigraph-core --lib` — 3 new regression tests:
  - `init_recovers_from_transient_open_failure_without_wiping` — a file that heals within the first retry window is not wiped
  - `init_recovers_from_slower_transient_failure_via_backoff` — heals at ~600ms, past where a single fixed-delay retry would already have given up, proving the backoff schedule (not just one retry) matters
  - `init_wipes_and_rebuilds_on_persistent_corruption` — durable corruption still correctly falls through to wipe+rebuild
- [x] Full `infigraph-core` test suite (271 tests) passes
- [x] `graph_queries` integration suite (63 tests) passes

### DIFF
--- a/crates/infigraph-core/src/lib.rs
+++ b/crates/infigraph-core/src/lib.rs
@@ -90,6 +90,10 @@ enum BackendKind {
 }
 
 impl Infigraph {
+    /// Backoff schedule (ms) for retrying a non-lock-contention graph open
+    /// failure before concluding it's genuine corruption. See `init()`.
+    const OPEN_RETRY_BACKOFF_MS: [u64; 3] = [200, 500, 1000];
+
     /// Open a project directory. Creates `.infigraph/` if it doesn't exist.
     pub fn open(root: &Path, registry: LanguageRegistry) -> Result<Self> {
         Self::open_shared(root, std::sync::Arc::new(registry))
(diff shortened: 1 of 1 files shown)


murari316 (approved):
Rebased onto current main (was based on stale main from before today's merges). Build/tests/fmt/clippy clean on rebased head. Investigated a watch/mod.rs symlink test failure that appeared during rebase testing (watch_project_detects_changes_through_symlinked_root) -- confirmed via repeated same-binary runs (no rebuild between) that it's a pre-existing flake unrelated to this PR's diff (which only touches Infigraph::init()'s retry/backoff logic in lib.rs), not a regression.

PR #25, opened 2026-07-20
feat: lock-file identity, bounded-wait acquisition, and Busy diagnostics

pradeepmouli:
## What
- New `infigraph-core::lockfile` module: JSON identity payload (`pid` / `role` / `build_hash` / `acquired_at`) stamped into held lock files; bounded-wait acquisition with 1ms→500ms exponential backoff; a structured `Busy` error naming the current holder; conservative stale handling — the kernel flock is the liveness truth (a free flock with a leftover payload is adopted; a held flock with an unreadable payload is an unknown holder and is never broken).
- `WriteLock` (the graph write lock at `<db>.lock`) now delegates to it. Public API unchanged; behavioral change: `write_lock()` waits up to 30s and then errors with a downcastable `Busy` instead of blocking forever. New `GraphStore::write_lock_with_timeout` for caller-chosen budgets.
- Compile-time build hash (`build.rs` → `INFIGRAPH_BUILD_HASH`, git-dir-aware rerun tracking) so a stale-binary lock holder is identifiable from the lock file alone.

## Why
Today the graph write lock is a 0-byte flock: a wedged or orphaned holder hangs every other process silently and undiagnosably. This makes contention observable (who holds it, since when, built from what) and bounded (structured timeout errors instead of infinite blocks). It is groundwork for follow-up PRs adding an operation-scoped index lock and shared-state (sessions/registry) locking.

## Compatibility
- flock semantics unchanged — old and new binaries interoperate on the same lock files; an old binary's bare flock is reported as "unknown holder" and never broken.
- All existing call sites compile unchanged; the pre-existing `write_lock{,_edge_cases,_wiring,_perf}` suites pass unmodified.

## Testing
- 15 new unit/integration tests (identity stamping, stale adoption, bounded-wait timeout with holder naming, unknown-holder conservatism, cross-handle contention), written TDD.
- Full `infigraph-core` suite green: 29 suites, 589 passed, 0 failed.
- Contended-throughput perf check (ignored/manual test): 3/3 runs within threshold.

https://claude.ai/code/session_01GhPBkDMHPpgcYJ3TKLAC1r

### DIFF
--- a/crates/infigraph-core/build.rs
+++ b/crates/infigraph-core/build.rs
@@ -1,0 +1,60 @@
+use std::path::Path;
+use std::process::Command;
+
+fn main() {
+    let sha = Command::new("git")
+        .args(["rev-parse", "--short=12", "HEAD"])
+        .output()
+        .ok()
+        .filter(|o| o.status.success())
+        .map(|o| String::from_utf8_lossy(&o.stdout).trim().to_string());
+
+    let dirty = Command::new("git")
+        .args(["status", "--porcelain"])
+        .output()
+        .ok()
+        .filter(|o| o.status.success())
+        .map(|o| !o.stdout.is_empty())
+        .unwrap_or(false);
+
+    let hash = match sha {
+        Some(s) if dirty => format!("{s}-dirty"),
+        Some(s) => s,
+        None => "unknown".to_string(),
+    };
+    println!("cargo:rustc-env=INFIGRAPH_BUILD_HASH={hash}");
+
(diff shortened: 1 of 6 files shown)


murari316 (approved):
Reviewed diff in full — lockfile.rs is well-structured (flock as liveness truth, identity payload for diagnostics only, bounded-wait exponential backoff, Busy error naming holder). 12 new tests cover identity roundtrip, stale-payload adoption, timeout naming, and old-binary compatibility. Approving.

PR #40, opened 2026-07-29
fix: SCIP symbol-name and enrichment correctness (parameter descriptor pollution, span/name/enrichment)

pradeepmouli:
## Summary

Three related fixes to SCIP import correctness in `crates/infigraph-core/src/scip/mod.rs`, found while investigating spurious/ugly entries in `get_doc_context` output:

1. **`scip_sym_to_name` returned empty/wrong names for most real SCIP symbols.** The old implementation split on suffix characters (`.`, `#`, `/`, `:`) without first stripping the trailing suffix marker itself, so it frequently returned an empty string or included the marker. Rewritten to strip the trailing terminator/disambiguator group first, then extract the name, with explicit handling for backtick-quoted descriptors (`` `test.ts`/Animal# ``) and bare identifiers.

2. **SCIP enrichment overwrote existing symbols' `start_line`/`end_line`.** A SCIP definition occurrence's range is the span of the identifier token itself, not the enclosing declaration — but enrichment was writing that narrow range back over the existing symbol's correct full-body span (already established by tree-sitter's AST extraction), silently collapsing every enriched symbol's displayed source to ~1 line. Enrichment now only ever updates `docstring`.

3. **Parameter/member descriptors were becoming spurious top-level graph symbols.** A simple function parameter's SCIP moniker is `<enclosing-method-moniker-ending-in-.>.(paramName)` — no `local ` prefix, so it bypassed the existing local-symbol filter and was created as its own (incorrectly named, since it depends on fix #1) graph node. Root-caused against real `@sourcegraph/scip-typescript` 0.4.0 output; `SymbolInformation.kind` was verified to always be `UnspecifiedKind` in real output, so a `kind`-based filter wasn't viable. Fixed with a structural check (`is_member_of_known_symbol`) that strips a trailing `(...)` descriptor group and checks whether what remains is already a known symbol from this same import pass — this exploits SCIP's own Parameter-descriptor grammar directly, so it should generalize across every SCIP indexer, not just TypeScript's.

Each fix is independently tested; #3 was verified to not affect CALLS-edge resolution (Pass 2) or INHERITS-edge resolution via relationships (Pass 3) — traced through both passes to confirm a suppressed parameter symbol cleanly fails to resolve as a reference target rather than dangling or misattributing an edge.

## Test plan

- [x] `cargo test -p infigraph-core --lib scip::` — 7/7 pass (all new + pre-existing tests)
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy -p infigraph-core --all-targets -- -D warnings` — clean
- [x] Verified end-to-end against a real indexed TypeScript sample

### DIFF
--- a/crates/infigraph-core/src/scip/mod.rs
+++ b/crates/infigraph-core/src/scip/mod.rs
@@ -105,7 +105,15 @@ pub fn import_scip_index(
     }
 
     // Pass 1: collect enrichments and new symbols in memory
-    let mut enrichments: Vec<(String, u32, u32, String)> = Vec::new();
+    //
+    // Enrichments only ever update `docstring`. Existing symbols already
+    // have a correct start_line/end_line from tree-sitter's full-body AST
+    // extraction; a SCIP definition occurrence's range is the span of the
+    // identifier token itself, not the enclosing declaration, so it must
+    // never be written back over the tree-sitter-derived span (previously
+    // it was, silently collapsing every enriched symbol's displayed source
+    // to ~1 line).
+    let mut enrichments: Vec<(String, String)> = Vec::new();
     let mut new_symbols: Vec<(String, String, String, String, u32, u32, String)> = Vec::new();
 
     for doc in &index.documents {
@@ -137,12 +145,7 @@ pub fn import_scip_index(
             let key = (file.clone(), name.clone());
             if let Some(ids) = file_name_to_ids.get(&key) {
                 for sid in ids {
-                    enrichments.push((
-                        sid.clone(),
-                        span.start_line,
-                        span.end_line,
-                        docstring.to_string(),
-                    ));
+                    enrichments.push((sid.clone(), docstring.to_string()));
                     stats.symbols_enriched += 1;
                 }
             } else {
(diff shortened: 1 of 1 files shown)


murari316:
Not applicable -- #40 is a PR, this should be a PR comment

murari316:
Closing in favor of a split. This PR's scip_sym_to_name rewrite (fix 1) conflicts with #44, which was merged first and whose version is a strict superset (it also handles chained disambiguator groups and quoted-descriptor digit-suffix disambiguators that motivated #44's own crash fix) -- merging #40 as-is would silently regress #44's fix via merge-conflict resolution.

Fixes 2 and 3 (SCIP enrichment must not overwrite tree-sitter's start_line/end_line, and parameter/member descriptors must not become spurious top-level graph symbols) are real, distinct, well-tested, and don't conflict with anything -- these have been rebased onto current main and merged separately, commit 6b9a14e (branch fix/scip-enrichment-span-and-param-suppression), with full attribution to this PR and its original author in the commit message.

Thanks for the thorough root-causing here -- the is_member_of_known_symbol structural check and the enrichment-span test in particular were exactly right and landed unchanged.

PR #54, opened 2026-08-07
Unify 5 hardcoded ignore-directory lists into a shared gitignore-aware component

pradeepmouli:
## Summary

Replaces 5 independently-hardcoded ignore-directory lists with a single shared `infigraph_core::ignore_rules` component that honors real `.gitignore` rules (via the `ignore` crate, ripgrep's crate) plus a custom `.infigraphignore` file with the same syntax — fixing CLAUDE.md's previously-false invariant that indexing already respects `.gitignore`/`.infigraphignore`.

Unified call sites:
- File/code discovery walker
- Document indexing (`infigraph-docs`)
- The file watcher (directory registration + event filtering) — uses a point-wise `IgnoreMatcher` since it evaluates single filesystem events rather than doing a directory walk
- `grep_search`
- Security scanning (`detect_security_issues`)

A fixed safety list (`.git`, `node_modules`, `target`, `.venv`, etc.) is still always excluded regardless of ignore files, on top of real `.gitignore`/`.infigraphignore` support.

**Motivation:** a live incident showed the doc-watch daemon stuck in an infinite "0 files reindexed" loop because a gitignored worktree-scratch directory (`scratchpad/`) wasn't in any of the 5 hardcoded lists, so it kept getting walked and watched.

## Notable implementation detail

`IgnoreMatcher::is_ignored()` manually walks ancestor directories because `ignore::Gitignore::matched()` only tests a path's own final segment against patterns — it does not automatically check ancestors. This was the crux of two fix rounds during implementation and was independently re-verified (hand-traced against the crate's real matching semantics) during final review.

## Test plan

Independently cherry-picked and re-verified on a fresh branch off real `upstream/main` (843eeb4), not carried over from the fork's own CI run, per this fork's standing "verify from a clean upstream base" convention:

- [x] `cargo build -p infigraph-core -p infigraph-cli -p infigraph-mcp` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo test -p infigraph-core --lib` — 308 passed (includes the 4 dedicated `ignore_rules` tests and the code-watcher's ignore coverage)
- [x] `cargo test -p infigraph-docs --lib --test modules` — 57 passed

Each of the 5 unified call sites has a dedicated test proving a previously-unreachable gitignored directory is now correctly excluded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ Cargo.lock
 
 # IDE / Editor
 .claude/
+.worktrees/
 .vscode/
 .idea/
 *.swp
--- a/crates/infigraph-core/src/ignore_rules.rs
+++ b/crates/infigraph-core/src/ignore_rules.rs
@@ -127,9 +127,28 @@ impl IgnoreMatcher {
         {
             return true;
         }
-        // The Gitignore was built for self.root, and matched() should handle
-        // paths that are within the root, whether absolute or relative
-        self.gitignore.matched(path, is_dir).is_ignore()
+        // Strip root prefix to get relative path for gitignore matching.
+        let rel_path = path.strip_prefix(&self.root).unwrap_or(path);
+
+        // Check if the path itself matches gitignore rules
+        if self.gitignore.matched(rel_path, is_dir).is_ignore() {
+            return true;
+        }
+
+        // Check if any parent directory matches gitignore rules (directories
+        // like "scratchpad/" in .gitignore should exclude all descendants)
+        let mut current = rel_path;
+        while let Some(parent) = current.parent() {
+            if parent == Path::new("") {
+                break;
+            }
+            if self.gitignore.matched(parent, true).is_ignore() {
+                return true;
+            }
+            current = parent;
+        }
+
+        false
     }
 }
 
--- a/crates/infigraph-core/src/ignore_rules.rs
+++ b/crates/infigraph-core/src/ignore_rules.rs
@@ -5,7 +5,6 @@
 //! directory-name list.
 
 use std::path::Path;
-use std::sync::Arc;
 
 use ignore::gitignore::{Gitignore, GitignoreBuilder};
 use ignore::WalkBuilder;
@@ -42,34 +41,12 @@ fn is_safety_excluded(name: &str) -> bool {
 /// configuration (e.g. `.max_depth`) before calling `.build()`.
 pub fn walk_builder(root: &Path) -> WalkBuilder {
     let mut builder = WalkBuilder::new(root);
-
-    // Discover and add all .gitignore and .infigraphignore files
-    let mut gi_builder = GitignoreBuilder::new(root);
-    let mut discovery = WalkBuilder::new(root);
-    discovery
-        .hidden(false)
+    builder
+        .hidden(true)
+        .git_ignore(true)
+        .require_git(false)
+        .add_custom_ignore_filename(".infigraphignore")
         .filter_entry(|entry| !is_safety_excluded(&entry.file_name().to_string_lossy()));
-
-    for result in discovery.build() {
-        let Ok(entry) = result else { continue };
-        let name = entry.file_name().to_string_lossy();
-        if name == ".gitignore" || name == ".infigraphignore" {
-            let _ = gi_builder.add(entry.path());
-        }
-    }
-
-    let gitignore = Arc::new(gi_builder.build().unwrap_or_else(|_| Gitignore::empty()));
-
-    // Apply gitignore rules via filter_entry
-    builder.hidden(true).filter_entry(move |entry| {
-        // Check safety list first
-        if is_safety_excluded(&entry.file_name().to_string_lossy()) {
-            return false;
-        }
-        // Then check gitignore rules
-        let is_dir = entry.file_type().is_some_and(|ft| ft.is_dir());
-        !gitignore.matched(entry.path(), is_dir).is_ignore()
-    });
     builder
 }
 
@@ -79,7 +56,6 @@ pub fn walk_builder(root: &Path) -> WalkBuilder {
 /// discover -- rebuild when those files may have changed (the watcher
 /// rebuilds this on its periodic tick; see `watch_project_with_periodic`).
 pub struct IgnoreMatcher {
-    #[allow(dead_code)]
     root: std::path::PathBuf,
     gitignore: Gitignore,
 }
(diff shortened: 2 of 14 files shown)


murari316 (changes requested):

PR 54 replaces five hardcoded ignore-directory lists with a shared gitignore/.infigraphignore-aware component (infigraph_core::ignore_rules) used by code discovery, doc indexing, the file watcher, grep_search, and security scanning, plus a fixed safety list. The direction is right and fixes a real incident (gitignored scratchpad/ looping the doc watcher), but the watcher-side IgnoreMatcher has two correctness defects (nested-.gitignore patterns mis-rooted; matcher never rebuilt in any production watcher despite the code comment claiming it is), the safety list silently un-indexes committed vendor/build/dist/coverage source with no override, and the headline end-to-end watcher regression test ships #[ignore]d so its new assertions never run in CI. Untouched invariants (advisory write lock, watcher lock file, kuzu open validation, wipe-on-open triage) are genuinely unaffected by this diff.

---

### [MAJOR] correctness — `crates/infigraph-core/src/ignore_rules.rs:77-91`
IgnoreMatcher::build adds every discovered nested .gitignore/.infigraphignore into a single GitignoreBuilder rooted at the project root, but the ignore crate interprets all patterns relative to the builder root (verified in ignore-0.4.30 source: GitignoreBuilder::add -> add_line uses the file path only as provenance). WalkBuilder builds a per-directory Gitignore, so the watcher's event filter and the indexing walker apply different rules for any nested ignore file.

_Failure scenario:_ sub/.gitignore contains 'data/': WalkBuilder ignores only sub/**/data/, but IgnoreMatcher treats the pattern as root-relative-unanchored and ignores other/data/** too, so the watcher drops events for files that ARE indexed -> permanently stale index for other/data. Conversely '/generated' in sub/.gitignore anchors to root/generated instead of sub/generated, so the watcher indexes sub/generated files the walker excludes. Negation (!) patterns are similarly mis-scoped.

### [MAJOR] correctness — `crates/infigraph-core/src/watch/mod.rs:139`
The IgnoreMatcher rebuild is gated on periodic_secs > 0, but both production entry points (crates/infigraph-mcp/src/tools/watch.rs:256 and crates/infigraph-cli/src/info_commands.rs:332) call watch_project, which hardcodes periodic_secs = 0 -- so the matcher is built once at watcher start and never refreshed in any real deployment. The doc comment in ignore_rules.rs ('the watcher rebuilds this on its periodic tick') is false for every production caller.

_Failure scenario:_ The incident's own remediation path breaks: a user adds 'scratchpad/' to .gitignore while the watcher is running (exactly how the 2026-08-06 incident would be fixed live). The stale matcher keeps passing scratchpad events, the watcher keeps indexing gitignored content until the watcher is manually restarted, and nothing signals staleness.

### [MAJOR] behavioral-regression — `crates/infigraph-core/src/ignore_rules.rs:17-33`
IGNORE_SAFETY_LIST is the union of all five old lists and is enforced via filter_entry, which runs before any ignore-file matching -- so vendor, build, dist, coverage, .idea, .mypy_cache, .pytest_cache are now unconditionally excluded from CODE indexing (collect_files at crates/infigraph-core/src/lib.rs:601 previously excluded only .infigraph/node_modules/__pycache__/.tox plus hidden), and neither .gitignore nor .infigraphignore whitelisting (!vendor) can resurrect them.

_Failure scenario:_ A Go repo with a committed vendor/ tree (standard 'go mod vendor' workflow: not hidden, not gitignored, previously indexed): after upgrading, reindex silently drops every symbol under vendor/ -- search/trace_callers return nothing for code that was queryable yesterday, with no error and no configuration escape hatch. Same for projects with a real build/ or coverage/ source directory.

### [MAJOR] test-coverage — `crates/infigraph-mcp/tests/watcher_reindex.rs:773`
The end-to-end watcher regression test for the incident scenario (test_code_watcher_ignores_excluded_dirs) is marked #[ignore] in this same PR (citing pre-existing issue #53), so the newly added scratchpad assertions never execute in CI. The watcher is the incident's actual code path, and its live-event behavior (IgnoreMatcher) is precisely where findings 1-2 live; only matcher-level unit tests in ignore_rules.rs run. This also contradicts the PR body's claim that each of the 5 unified call sites has a dedicated (running) test.

_Failure scenario:_ A future change breaks IgnoreMatcher event filtering (e.g. either bug above, or a refactor of is_ignored); CI stays green on all three platforms because the only test exercising watcher-event ignore behavior is skipped, and the incident recurs.

### [MINOR] correctness — `crates/infigraph-core/src/ignore_rules.rs:103-117`
IgnoreMatcher::is_ignored checks only the safety list and gitignore rules -- it drops the old watcher rule that any dot-prefixed path component is ignored (old should_ignore, watch/mod.rs). walk_builder uses hidden(true), so the watcher now accepts events for hidden files/dirs the walker would never index.

_Failure scenario:_ Mid-watch, a tool creates .backup/copy.py or a user saves .scratch.py: the Created event passes is_ignored, the .py extension matches the registry, and the file is indexed (new hidden dirs are even registered for watching via register_watch_dirs). The next full reindex drops it -- the index flip-flops between watcher and walker views.

### [MINOR] correctness — `crates/infigraph-core/src/ignore_rules.rs:42-49`
walk_builder inherits WalkBuilder defaults it never pins: .ignore files, global gitignore (~/.config/git/ignore), .git/info/exclude, and parent-directory ignore files all apply to walking, but IgnoreMatcher::build collects only .gitignore/.infigraphignore under root -- another walker/matcher divergence, plus per-machine nondeterminism in what gets indexed.

_Failure scenario:_ A rule in .git/info/exclude (or a developer's global gitignore, e.g. '*.gen.ts') excludes files from indexing on one machine but the watcher still processes their events; or two teammates index the same repo and get different graphs because of their personal global gitignore.

### [MINOR] error-handling — `crates/infigraph-core/src/security/detect.rs:33-37`
walk_and_scan changed from propagating std::fs::read_dir(dir)? to swallowing all walk errors (Err(_) => continue), so scan_project on a nonexistent or unreadable root now returns Ok with zero findings instead of an error. Same silent-empty pattern introduced in DocIndex::collect_doc_files (crates/infigraph-docs/src/lib.rs:318).

_Failure scenario:_ detect_security_issues is pointed at a mistyped or permission-denied path: it reports a clean scan (0 findings) rather than failing, and a consumer treats the project as vetted.

### [MINOR] behavioral-regression — `crates/infigraph-core/src/security/detect.rs:27-40`
Security scanning now skips gitignored files and hidden files (walk_builder: hidden(true) + git_ignore). The old scanner skipped only hidden/hardcoded directories -- individual files, including gitignored ones like config/local_settings.py, were scanned. Gitignored files are exactly where hardcoded secrets tend to live, so applying indexing's ignore semantics to a security scanner deserves an explicit decision, not an incidental unification.

_Failure scenario:_ config/secrets.py containing a hardcoded credential is gitignored (as intended); before this PR detect_security_issues flagged it, after this PR it is invisible to the scan.

### [NIT] correctness — `crates/infigraph-core/src/watch/mod.rs:215`
is_ignored(&path, path.is_dir()) evaluates is_dir on a possibly-deleted path: for a Deleted event on an ignored directory itself, is_dir() is false so a dir-only pattern like 'scratchpad/' doesn't match the path (ancestor check only covers parents), and the deletion event leaks through the filter.

_Failure scenario:_ rm -rf scratchpad/ while watching: the watcher processes a spurious Deleted event for a directory that was never indexed -- harmless removal work today, but a trap if the Deleted branch ever grows side effects.

### [NIT] error-handling — `crates/infigraph-core/src/ignore_rules.rs:95`
gi_builder.build().unwrap_or_else(|_| Gitignore::empty()) silently degrades to an empty matcher on build failure, reverting the watcher to pre-fix (incident) behavior with no log line; likewise let _ = gi_builder.add(...) drops file-read errors silently.

_Failure scenario:_ A malformed .gitignore causes build() to fail; the watcher silently stops honoring every ignore rule except the safety list and re-enters the exact 'walk gitignored scratch dirs' failure this PR exists to fix, with nothing in the watch log to explain it.

PR #61, opened 2026-08-13
fix: guard release tag against workspace version mismatch

### LINKED ISSUE #49: infigraph --version reports 3.2.6 in both v3.2.7 and v3.2.8 release binaries
## Summary

`infigraph --version` reports `infigraph 3.2.6` even when running the actual v3.2.7 or v3.2.8 release binary. `infigraph update` correctly detects and installs the newer release (binary hash changes, confirming real code was shipped), but the CLI's own `--version` output never advances past `3.2.6`.

## Repro

```
$ infigraph update
Updating infigraph: v3.2.6 → v3.2.8
Downloading infigraph-aarch64-apple-darwin.tar.gz from release v3.2.8...
Installed v3.2.8 to /Users/mlal/.local/bin
...
$ infigraph --version
infigraph 3.2.6
```

Confirmed this is not a stale-binary/PATH/caching issue on the local machine — downloaded each darwin release asset fresh from GitHub and ran `--version` directly:

| Release tag | SHA-256 (aarch64-apple-darwin) | `--version` output |
|---|---|---|
| v3.2.5 | (not recorded) | `infigraph 3.2.5` (correct) |
| v3.2.6 | `a1019bb2...` | `infigraph 3.2.6` (correct) |
| v3.2.7 | `4adb0943...` | `infigraph 3.2.6` (**stale**) |
| v3.2.8 | `a1019bb2...` | `infigraph 3.2.6` (**stale**) |

Note v3.2.6 and v3.2.8 share an identical binary hash for this platform, which is separately suspicious (either an accidental no-op release for aarch64-darwin, or a build artifact reuse bug) — but even the v3.2.7 binary, which does have a distinct hash from v3.2.6 (i.e. new code was actually compiled and shipped), still reports the old version string.

## Root cause (suspected, not yet located)

The version string returned by `--version` (likely `CARGO_PKG_VERSION` via clap, or a hardcoded constant) was not bumped in the source/`Cargo.toml` before cutting the v3.2.7 and v3.2.8 releases, even though other code changes did go in. This means:
- Anyone scripting against `infigraph --version` to gate behavior on a minimum version gets a false negative for two releases running.
- It's impossible to tell from `--version` alone whether you're actually on v3.2.7/v3.2.8 functionality or still on v3.2.6.

## Impact

Users (and `infigraph update` callers) cannot trust `--version` to confirm they're on the release they just installed. Combined with the v3.2.6/v3.2.8 aarch64-darwin hash collision, this also raises the question of whether the v3.2.8 release for this platform actually contains the intended changes.

## Suggested fix

- Find wherever the CLI's version constant is set (Cargo.toml `version`, or a hardcoded string passed to clap's `Command::version(...)`) and confirm the release pipeline actually bumps it before building/tagging.
- Add a release-CI check that fails the release if the built binary's `--version` output doesn't match the git tag being published.
- Separately verify why v3.2.6 and v3.2.8's aarch64-apple-darwin binaries are byte-identical — confirm the v3.2.8 release actually contains new code for this platform target.

Atishyy27:
## Summary

Fixes #49.

`release.sh` takes the release tag as an arbitrary CLI arg and never checks it against `[workspace.package].version` in `Cargo.toml`. `infigraph-cli`'s `--version` comes from clap's `version` field, which pulls `CARGO_PKG_VERSION` at compile time from that same `Cargo.toml`. So a forgotten version bump (or a typo in the tag arg) ships binaries whose `--version` output silently disagrees with the git tag/GitHub release they're uploaded under, with no way to fix it after the fact short of deleting the release — matching exactly what #49 reports (`--version` stuck on 3.2.6 across the v3.2.7 and v3.2.8 releases).

Guards by comparing the tag (v-prefix stripped) against the workspace version before any build starts, and fails with a clear message pointing at the fix in either direction.

## Test plan

- [x] `bash -n release.sh` passes (syntax)
- [x] Tested the guard logic in isolation against this repo's real `Cargo.toml` (currently `3.2.15`): a mismatched tag (`v9.9.9`) fails with the new error message and exit 1; a matching tag (`v3.2.15`) passes through cleanly
- [ ] `cargo test --all` / `cargo clippy` — not applicable, this is a shell-only change, no Rust touched

## Notes

Didn't run the full release build (requires cmake + a real GitHub release target), the change is isolated to the version-guard block before any build/upload step runs, so it can't affect build/package/upload behavior when the versions do match.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -23,6 +23,17 @@ fi
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 cd "$SCRIPT_DIR"
 
+# Guard: the tag must match the workspace version, or the binaries built
+# below will report a version (via clap's `version` from CARGO_PKG_VERSION)
+# that disagrees with the release tag they're uploaded under.
+CARGO_VERSION="$(grep -m1 '^version = ' Cargo.toml | sed -E 's/version = "(.*)"/\1/')"
+TAG_VERSION="${VERSION#v}"
+if [ "$TAG_VERSION" != "$CARGO_VERSION" ]; then
+  echo "Error: release tag ${VERSION} does not match [workspace.package].version ${CARGO_VERSION} in Cargo.toml."
+  echo "Bump the workspace version to ${TAG_VERSION}, or run: $0 v${CARGO_VERSION}"
+  exit 1
+fi
+
 # Detect current platform
 OS="$(uname -s)"
 


Atishyy27:
Duplicate of #60, opened by mistake within minutes of it (an earlier session's work I didn't check for before re-diagnosing and re-fixing the same issue). Closing this one, #60 has the fix.

Issue #50, opened 2026-08-03
Default installation is too invasive


kostis-codefresh:
I tried infigraph for over 2 weeks. Initially, I liked it, but I completely uninstalled it as the gains were not worth the effort.

The default installation is too invasive for my local Claude setup.

1. It creates several hooks that run on their own
2. It created autogenerated CLAUDE.MD files 
3. It constantly opens/closes/records search sessions

My original use for infinigraph was to find something better than grep/sed/awk/lsp and also save potentially on tokens. But it turned out to be the exact opposite. On some of my sessions, 30% of usage was just on Infigraph

I understand what you are trying to do, but at this point in time I believe a simpler installation mode is needed. 

1. The simplest mode would be just the mcp server and nothing else. I would like to call it manually and use the killer feature of finding the blast radius of a change.
2. Then the second mode, where things are mostly automated for indexing and searching
3. Then a more complex mode where you record sessions, extract statistics, etc.

People should be able to choose mode upon installation.

Right now I cannot use infigraph because it tries to do too much and makes the experience even worse.

PR #63, opened 2026-08-19
Data-driven artifact engine for agent integrations (fixes #29, groundwork for #50)

### LINKED ISSUE #29: `infigraph install` writes incompatible MCP configurations for some coding agents
`infigraph install` uses the same `mcpServers` JSON structure for several coding agents. However, some agents require different configuration paths or schemas.

I manually reproduced this with GitHub Copilot CLI and OpenCode. I also found possible mismatches for other integrations by comparing `config_targets.rs` with their official documentation.

## Related issue

The equivalent Codex TOML configuration problem is already tracked in #26 and addressed by PR #27.

Codex is excluded here. This issue focuses on the shared JSON configuration writer and JSON-based integrations.

### GitHub Copilot CLI

Infigraph writes:

```text
~/.copilot/mcp.json
```

Copilot CLI expects:

```text
~/.copilot/mcp-config.json
```

`copilot mcp list` did not detect Infigraph at the generated path. It was detected after placing the configuration at the documented path.

Documentation:  
https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/add-mcp-servers

### OpenCode

Infigraph writes:

```text
~/.opencode/config.json
```

using a top-level `mcpServers` object.

OpenCode expects:

```text
~/.config/opencode/opencode.json
```

using the following structure:

```json
{
  "mcp": {
    "infigraph": {
      "type": "local",
      "command": ["/path/to/infigraph-mcp", "--mcp"]
    }
  }
}
```

OpenCode discovered and attempted to start Infigraph after using the documented path and schema.

Documentation:

- https://opencode.ai/docs/config/
- https://opencode.ai/docs/mcp-servers/

## Other documentation mismatches

The following integrations were compared against their documentation but were not manually tested by me:

| Agent | Infigraph writes | Documentation suggests |
|---|---|---|
| VS Code | `~/.vscode/mcp.json` with `mcpServers` | User/workspace MCP configuration using `servers` |
| Zed | `~/.zed/mcp.json` with `mcpServers` | Zed settings using `context_servers` |
| Windsurf | `~/.windsurf/mcp.json` | `~/.codeium/windsurf/mcp_config.json` |
| Kiro | `~/.kiro/mcp.json` | `~/.kiro/settings/mcp.json` |

Documentation:

- VS Code: https://code.visualstudio.com/docs/agent-customization/mcp-servers
- Zed: https://zed.dev/docs/ai/mcp
- Windsurf: https://docs.windsurf.com/windsurf/cascade/mcp
- Kiro: https://kiro.dev/docs/mcp/configuration/

Kiro requires additional verification because some of its documentation still mentions the older `~/.kiro/mcp.json` path.

Cursor and Gemini appear to use the paths and `mcpServers` structure currently generated by Infigraph.

## Root cause

All JSON-based targets use `install_json_target()`, which always writes an `mcpServers` object.

This assumes that every JSON-based coding agent uses the same path and schema.

## Expected behavior

`infigraph install` should select the correct path and configuration writer for each supported agent.

Target-specific tests should verify:

- The documented path and schema
- Existing settings and other MCP servers are preserved
- Running installation repeatedly does not create duplicates
- Uninstallation removes only Infigraph

## Environment

```text
Infigraph version: 3.2.5
Rust version: rustc 1.97.1
macOS version: 15.2
Manually tested: GitHub Copilot CLI and OpenCode
Documentation comparison only: VS Code, Zed, Windsurf and Kiro
```

pradeepmouli:
## Summary

Replaces the hardcoded per-agent install logic (`config_targets.rs`'s `AGENT_TARGETS` array, ~10 `install_*_hook` functions, and the ad-hoc `write_claude_md_instructions`/`write_editor_rules`/`write_reindex_command` functions) with a data-driven **artifact engine**:

- Every agent integration is a directory under `crates/infigraph-cli/resources/integrations/<agent>/` (bundled into the binary via `build.rs`), optionally overridden per-file at `~/.infigraph/integrations/<agent>/` — same layout, user wins.
- Most artifacts need zero manifest entries: `.json` files convention-infer `json_deep_merge`, everything else non-`.toml` convention-infers `overwrite`. `.toml` destinations always need an explicit `config.toml` manifest entry (`toml_section` + `key_path`), since a full parse/reserialize would risk stripping comments from a hand-maintained file like Codex's `~/.codex/config.toml`.
- Five apply/remove strategies: `json_deep_merge`, `overwrite`, `marker_delimited`, `toml_section`, `json_key_path`.
- VS Code and Zed (OS/profile-dependent destination paths) use a resolver escape hatch — a subprocess with a JSON stdin/stdout contract, same shape as `infigraph-pipeline-plugin`'s `command` field: `resolver = ["python3", "./resolve-x-path.py"]`, full command array, author-controlled interpreter.
- `cmd_install`/`cmd_uninstall` are now thin wrappers around `discover_artifacts` + `apply_resolved_artifact`/`remove_resolved_artifact`; uninstall symmetry is derived from each artifact's own shape rather than hardcoded per-agent keys.
- New doc: `docs/AGENT_INTEGRATION_PLUGINS.md`, written in the same format as `GRAMMAR_PLUGINS.md`/`docs/PIPELINE_PLUGINS.md`.

Fixes #29 (wrong MCP config shape for OpenCode/Zed/Copilot CLI/Windsurf/Kiro/VS Code — each now gets its own correct schema and path instead of one hardcoded `mcpServers` shape). Lays groundwork for #50 (invasive default install): artifacts are now classified via an `InstallStep` enum (`McpRegistration`/`DocsAndRules`/`Hooks`/`Models`), grouping what `cmd_install` writes so a future `--mode` flag has something to filter on — no such flag ships in this PR.

Also folds in two small, independently-verified fixes found along the way (not scoped to this PR but touching the same install path): the `--mcp` flag alone triggering the web UI keep-alive loop, and a missing `type`/`tools` field pair in the bundled GitHub Copilot CLI MCP config.

## Safety fixes from an adversarial review pass

Before opening this PR, I ran an adversarial review against the full diff and verified two findings against the actual source before fixing them:

1. **Uninstall could silently delete a hand-edited file.** The `overwrite` strategy (hook scripts, Cursor/Windsurf rule files, shared skill docs) removed whatever was at its target path unconditionally — no different from the pre-existing `hooks.rs` behavior for hook scripts, but now covering a much larger surface (rule/skill files a user is far more likely to hand-edit). `overwrite` now records a content hash at install time (`~/.infigraph/installed-files.json`) and checks it before deleting; a changed file is left in place and reported instead of removed. The other four strategies don't need this — they already scope removal to a specific key/section/marker, so user edits elsewhere in the same file are untouched by construction.
2. **A `Skipped` outcome's `manual_snippet` was computed but discarded.** `apply_json_deep_merge`/`apply_json_key_path` compute the exact fragment to add by hand when a target file fails to parse (e.g. Zed's `settings.json` has comments or trailing commas), but `run_install`'s `Skipped { reason, .. }` match arm dropped it, so the user got a skip reason with no way to actually complete the registration. `InstallReport.skipped` now carries the snippet through to `cmd_install`'s output.

(A third finding from the same review — an OpenCode config schema claim — was checked against OpenCode's live docs and found to be incorrect; the shipped `mcp.<name>` shape is current.)

## Test plan

- [x] `cargo test --all`, `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings` all green
- [x] 153 tests in `infigraph-cli` cover the artifact engine end-to-end: bundled-fixture application per agent, uninstall symmetry, resolver dispatch (VS Code/Zed), and a new regression test asserting a hand-edited `overwrite` file survives uninstall while an unmodified one is removed
- [x] Manually verified `infigraph install`/`infigraph uninstall` against a scratch `$HOME`


murari316 (changes requested):

Replaces ~2000 lines of hardcoded per-agent install logic with a data-driven artifact engine: integrations are bundled resource directories (compiled in via build.rs, overridable under ~/.infigraph/integrations/) applied via five declarative strategies, with resolver subprocesses for OS-dependent paths (VS Code, Zed). The design is sound, test coverage is broad (per-agent fixtures, idempotency, uninstall symmetry), and the two folded-in fixes (--mcp UI keep-alive, Copilot CLI config shape) are unit-tested. However, the install/uninstall loops have no per-artifact fault isolation — a resolver spawn failure (no python3, common on Windows) or one malformed user-override manifest aborts the entire run before any MCP registration is written — and the advertised hand-edit preservation on uninstall is one-shot because the ownership entry is cleared even when preservation triggers. Repo invariants around the graph DB advisory lock, kuzu path validation, watcher lock, and wipe-on-corruption are genuinely untouched by this PR.

---

### [MAJOR] error-handling — `crates/infigraph-cli/src/install.rs:~105 (run_install apply loop) and ~296 (run_uninstall remove loop)`
No per-artifact fault isolation: run_install and run_uninstall propagate any single artifact failure with `?`, and apply_resolved_artifact/remove_resolved_artifact turn a resolver spawn failure (resolver.rs:57, run_resolver spawn) into a hard error. The bundled VS Code and Zed integrations declare `resolver = ["python3", ...]`, creating a new hard runtime dependency on python that the old install path never had. Because manifest-driven (pass-1) artifacts apply before all pass-2 convention artifacts, and every MCP-registration JSON (.claude.json, gemini, cursor, kiro, settings.json, all hooks) is pass-2, the failure lands before a single MCP registration is written. The same root cause means one malformed user-override config.toml (discovery.rs parse `?`) bricks both install and uninstall for every integration. Fix: map resolver spawn failure to ApplyOutcome::Skipped with the manual snippet, and collect per-artifact errors in the loops instead of aborting.

_Failure scenario:_ Windows or minimal-Linux machine without python on PATH: `infigraph install` writes CLAUDE.md and a couple of rule files, then dies at the vscode resolver — zero MCP servers registered, partial state left behind. Separately: a user with a typo in ~/.infigraph/integrations/foo/config.toml can never run `infigraph uninstall` — discovery errors out before removing anything.

### [MAJOR] correctness — `crates/infigraph-cli/src/artifacts/ownership.rs:66-72`
verify_unchanged_and_clear removes the manifest entry and saves the manifest BEFORE comparing hashes, so the hand-edit protection (the PR's advertised safety fix #1) is one-shot: after an uninstall that correctly preserves a modified file, the recorded hash is gone, and any subsequent uninstall treats the file as unrecorded ("safe to delete") and deletes it. Fix: keep the entry (or clear only on the return-true paths) when the hash mismatches.

_Failure scenario:_ User hand-edits ~/.cursor/rules/infigraph.mdc after install; runs `infigraph uninstall` — file preserved and reported; re-runs uninstall (e.g. to clear the remaining 'preserved' warnings, or out of habit) — remove_resolved_artifact now finds no manifest entry, verify returns true, and remove_overwrite silently deletes the hand-edited file.

### [MINOR] correctness — `crates/infigraph-cli/src/artifacts/strategy.rs:51 (is_owned_array_entry)`
Array-entry ownership is decided by the substring "infigraph" anywhere in the entry's serialized JSON. Both merge_json (on every install/upgrade) and remove_json_keys (on uninstall) will silently delete a user's OWN hook entries in ~/.claude/settings.json that merely mention infigraph — e.g. a personal wrapper hook script named my-infigraph-notifier.sh, or a hook whose matcher lists mcp__infigraph__* tools.

_Failure scenario:_ User adds their own PostToolUse hook `{"matcher": "mcp__infigraph__index_project", "hooks": [{"command": "~/bin/notify-reindex.sh"}]}`; the next `infigraph install` (or self-update's reinstall_hooks) filters it out as 'owned' and it vanishes without any message.

### [MINOR] correctness — `crates/infigraph-cli/src/artifacts/convention.rs:24 (catch-all `_ => Some(ConventionStrategy::Overwrite)`)`
Every unknown extension in the user-override directory becomes an installable overwrite artifact: editor droppings like foo.sh.bak, .infigraph-enforce.sh.swp, or file.json.orig in ~/.infigraph/integrations/<agent>/ are stripped of the agent prefix and written into $HOME (and recorded in installed-files.json). A modest denylist (bak/swp/tmp/orig) or an allowlist would prevent junk landing in the user's home.

_Failure scenario:_ User edits an override hook with vim, leaving claude-code/.claude/hooks/.infigraph-enforce.sh.swp behind; the next `infigraph install` writes ~/.claude/hooks/.infigraph-enforce.sh.swp into the user's real hooks directory.

### [MINOR] consistency — `crates/infigraph-cli/resources/integrations/claude-code/.claude/hooks/infigraph-enforce.sh:~120-135 (Read extension allowlist)`
Two artifacts shipped in this same PR give the model contradictory directives: shared/agents.md rule 6 says "Read files directly only for non-code files (configs, docs, manifests)", but the enforce hook's Read case blocks .md, .json, .toml, .yml, .ini, .conf, .cfg — exactly the config/doc file classes the instructions direct the model to Read. The model is denied for following the shipped instructions.

_Failure scenario:_ Claude, following the installed CLAUDE.md rules, calls Read on the project's Cargo.toml or README.md to check a manifest/doc; the enforce hook denies it and points at get_doc_context, which is a code-symbol tool and returns nothing useful for a manifest — a dead-end loop the sentinel escape hatch has to break every time.

### [MINOR] test-coverage — `crates/infigraph-cli/src/install.rs:tests module (cmd_install_writes_every_convention_based_integration_under_a_fake_home)`
The PR's advertised fix #2 — carrying manual_snippet through a Skipped outcome into InstallReport.skipped so users can complete a registration by hand — has no test at the run_install level. The only skipped-related assertion is `report.skipped.is_empty()` against an empty $HOME; strategy-level tests check ApplyOutcome::Skipped's snippet but nothing proves run_install propagates it rather than dropping it again (the exact regression being fixed).

_Failure scenario:_ A future refactor of run_install's match arm reverts to `Skipped { reason, .. }` (the original bug); all 153 tests stay green while users with a JSONC-commented Zed settings.json once again get a skip reason with no snippet.

### [NIT] simplification — `crates/infigraph-cli/src/install.rs:~135 (configured_integration_labels)`
configured_integration_labels re-runs the full discover_artifacts (re-reading the user override dir and re-parsing every manifest) just to map written paths back to labels; run_install already has the artifacts and could return labels in the report, avoiding the duplicate discovery and the possibility of the two passes disagreeing if the override dir changes between them.

### [NIT] consistency — `crates/infigraph-cli/src/artifacts/discovery.rs:343 (pass-2 `InstallStep::classify(relative_path, strategy)`)`
Convention-based artifacts pass the bundled SOURCE path (e.g. "claude-code/.claude/hooks/x.sh") to InstallStep::classify, contradicting classify's own doc comment ("by its destination path ... not the bundled-resources source path"). Harmless today because source mirrors destination plus a prefix, but an integration directory ever named "hooks", "rules", or "skills" would misclassify every artifact in it; the stripped target_relative_path computed two lines below is the correct argument.

PR #68, opened 2026-08-25
fix(graph): stop reusing a connection across independent bulk COPY writes

### LINKED ISSUE #67: Connection/transaction lifecycle bugs in KuzuBackend: wedged COPY retries + silently non-atomic writes
## Summary

Two real bugs in the Kùzu local backend share one root cause: `GraphStore` hands out bare `Connection` values and leaves every caller to manage connection lifetime and transaction boundaries by hand. Some callers reuse one connection across independent statements that never needed to share state (wedging a later, unrelated write); another set of callers believes it's inside a real transaction when it isn't, because of how `raw_query` handles transaction-control statements.

## Bug 1: connection reuse wedges a later, unrelated COPY

**Observed:** a `COPY Symbol` bulk load retries a few times after bad-primary-key errors (expected — drops the offending rows, retries). Immediately after, on the *same* connection, a `COPY CALLS` bulk load fails on its first attempt with Kùzu's internal `Invalid transaction type to rollback.` and falls back to the slower per-row `UNWIND` path. Self-healing (UNWIND produces correct data), but real and reproducible.

**Root cause:** `import_scip_index` (`crates/infigraph-core/src/scip/mod.rs`) opens one `Connection` at the top of the function and reuses it across an inline Symbol-node `COPY` block, then `copy_edges_with_bad_record_retry` (`crates/infigraph-core/src/graph/store_util.rs`) for `CALLS` edges, then again for `INHERITS` edges. None of these three bulk loads are wrapped in an explicit `BEGIN TRANSACTION` — each `COPY` already auto-commits independently regardless of connection sharing, so sharing the connection buys zero atomicity here while creating exposure to whatever internal state a caught COPY failure leaves behind. Same helper is also called from `resolve/calls.rs` and `store_parquet.rs`'s full-reindex bulk path.

**Proposed fix:** `copy_edges_with_bad_record_retry` should take `&GraphStore` instead of `&Connection`, calling `store.connection()?` fresh and unconditionally on every retry-loop iteration and before the `UNWIND` fallback — no "is this a retry" bookkeeping needed, since `GraphStore::connection()` already mints a fresh `Connection::new(&self.db)` cheaply on every call. The inline Symbol-`COPY` block in `import_scip_index` needs the same treatment.

## Bug 2: `raw_query`'s transaction no-op silently breaks real atomicity elsewhere

`KuzuBackend::raw_query` opens a fresh connection per call, so it deliberately no-ops `BEGIN TRANSACTION`/`BEGIN`/`COMMIT`/`ROLLBACK` when passed as a query string (intentional — fails safe rather than erroring on a stray `COMMIT`). Two call sites don't know this and rely on it for real atomicity:

- `concerns/mod.rs::write_concerns`: `BEGIN TRANSACTION`, `DETACH DELETE` all existing `Concern` nodes, loop `CREATE`-ing new ones, `COMMIT` — believing this is one atomic unit.
- `reflection/mod.rs`: same shape.

Because `raw_query`'s `BEGIN`/`COMMIT` are no-ops, every statement in these loops auto-commits independently and immediately. **A crash mid-loop today already means the old data is gone and only part of the new data landed** — live data-loss exposure, currently masked because the no-op returns `Ok` instead of erroring loudly.

**Proposed fix:** introduce `GraphStore::transaction<T>(&self, f: impl FnOnce(&Connection) -> Result<T>) -> Result<T>` — opens one connection, issues a real `BEGIN TRANSACTION`, runs the closure against that connection, commits on `Ok`, rolls back on `Err`. Migrate `write_concerns`, `reflection/mod.rs`'s equivalent, `KuzuBackend::write_calls_service_edges` (already correct today but duplicates the boilerplate), and the two hand-rolled transaction blocks in `KuzuBackend`'s bulk-index path onto it.

## Suggested tests

- Force a bad-PK retry on one table's COPY immediately followed by a COPY on a different table on what would previously have been the same connection — assert the second COPY succeeds cleanly.
- Inject a mid-loop failure into `write_concerns` and assert old `Concern` nodes are still present afterward (this should *fail* against current code, proving the live bug).
- `transaction()` itself: commit-on-success, rollback-on-error, and no stuck `BEGIN` on panic.

Happy to open a PR for this if useful.

pradeepmouli:
Fixes intuit/infigraph#67 (Bug 1 of 2 described there — the connection-reuse wedging).

## What

`copy_edges_with_bad_record_retry` took a borrowed `Connection` and reused it across every retry attempt and the `UNWIND` fallback. None of these bulk loads are wrapped in an explicit transaction, so sharing the connection bought no atomicity — it only created exposure to whatever internal state a caught `COPY` failure leaves behind.

Observed in production: a `Symbol`-table `COPY`'s bad-PK-drop-and-retry cycle left the connection wedged such that the very next `COPY` on it (a different table, `CALLS`) failed immediately with Kùzu's internal `Invalid transaction type to rollback.` and fell back to the slower per-row `UNWIND` path. Self-healing (identical output either way, per `store_bench::test_parquet_quality`), so no data-loss exposure — but real and reproducible.

## Fix

`copy_edges_with_bad_record_retry` now takes `&GraphStore` instead of `&Connection` and asks for a fresh connection on every retry-loop iteration and before the `UNWIND` fallback — no "is this a retry" bookkeeping needed, since `GraphStore::connection()` already mints a fresh, cheap `Connection` on every call. The inline Symbol-node `COPY`-with-retry block in `import_scip_index` (a near-duplicate of the same pattern for nodes instead of edges) gets the same treatment.

Threading `&GraphStore` down to the one caller of this helper that didn't already have it (`resolve_with_map`) also let `resolve_inherits` drop its now-entirely-unused `&Connection` parameter.

Also silences one pre-existing, unrelated `clippy::chunks_exact_to_as_chunks` lint in `embed/mod.rs` that the workspace-wide pre-commit hook caught (newer clippy than this branch's last-tested baseline) — no behavior change, matches clippy's own suggested suppression.

## Testing

- `cargo test -p infigraph-core --lib` (310 tests, all pass)
- `cargo test -p infigraph-core --test resolve_calls` (15 tests, all pass — exercises `resolve_inherits`/`resolve_with_map`)
- `cargo test -p infigraph-core --test graph_queries` (`_conn`-suffixed tests, all pass)
- `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` both clean
- Pre-commit hook's perf suite (`write_lock_perf`, `groups_watch_perf`, `index_perf`) all pass

Bug 2 from intuit/infigraph#67 (the `raw_query` transaction no-op silently breaking `write_concerns`/`reflection` atomicity) is a separate, larger change (introducing `GraphStore::transaction()`) and will follow as its own PR.

### DIFF
--- a/crates/infigraph-core/src/embed/mod.rs
+++ b/crates/infigraph-core/src/embed/mod.rs
@@ -398,6 +398,7 @@ pub fn load_embeddings(path: &Path) -> Result<Vec<(String, Vec<f32>)>> {
         pos += 4;
         let float_bytes = dim * 4;
         anyhow::ensure!(pos + float_bytes <= data.len(), "truncated embeddings file");
+        #[allow(clippy::chunks_exact_to_as_chunks)]
         let vec: Vec<f32> = data[pos..pos + float_bytes]
             .chunks_exact(4)
             .map(|chunk| f32::from_le_bytes(chunk.try_into().unwrap()))
--- a/crates/infigraph-core/src/graph/store_parquet.rs
+++ b/crates/infigraph-core/src/graph/store_parquet.rs
@@ -554,13 +554,13 @@ impl GraphStore {
             }
             let edge_pq = tmp.join(format!("infigraph_index_{}.parquet", table.to_lowercase()));
             copy_edges_with_bad_record_retry(
-                conn,
+                self,
                 table,
                 pairs.to_vec(),
                 src_label,
                 dst_label,
                 &edge_pq,
-            );
+            )?;
         }
 
         // Custom edge tables
@@ -574,13 +574,13 @@ impl GraphStore {
                 edge_name.to_lowercase()
             ));
             copy_edges_with_bad_record_retry(
-                conn,
+                self,
                 edge_name,
                 pairs.to_vec(),
                 "Symbol",
                 "Symbol",
                 &edge_pq,
-            );
+            )?;
         }
 
         // Cleanup node parquet files
--- a/crates/infigraph-core/src/graph/store_util.rs
+++ b/crates/infigraph-core/src/graph/store_util.rs
@@ -1,8 +1,10 @@
 use std::path::Path;
 
+use anyhow::Result;
 use kuzu::Connection;
 
 use crate::graph::parquet_loader;
+use crate::graph::store::GraphStore;
 
 /// How many bad-record-drop-and-retry cycles a bulk COPY gets before giving
 /// up and falling back to the slow per-row UNWIND path for whatever remains.
@@ -184,18 +186,30 @@ pub(crate) fn extract_bad_copy_value(err: &str) -> Option<&str> {
 /// retries are exhausted (`MAX_BAD_RECORD_RETRIES`) or the error isn't
 /// recognized as recoverable -- avoids paying the slow UNWIND path for an
 /// entire batch over a handful of bad rows.
+///
+/// Takes `store` rather than a borrowed `Connection`: a caught COPY failure
+/// can leave Kùzu's internal transaction bookkeeping on that connection in a
+/// state where the *next* statement fails immediately with `Invalid
+/// transaction type to rollback.` (observed in production -- a Symbol-table
+/// COPY's bad-PK retries left the connection wedged for the CALLS-table COPY
+/// that followed on the same connection). None of these bulk loads are
+/// wrapped in an explicit transaction, so sharing a connection across them
+/// buys no atomicity -- asking `store` for a fresh one every attempt is
+/// free of that risk and no more expensive (`GraphStore::connection` is a
+/// cheap `Connection::new` per call).
 pub(crate) fn copy_edges_with_bad_record_retry(
-    conn: &Connection,
+    store: &GraphStore,
     table: &str,
     mut pairs: Vec<(String, String)>,
     src_label: &str,
     dst_label: &str,
     edge_pq: &Path,
-) {
+) -> Result<()> {
     for attempt in 0..MAX_BAD_RECORD_RETRIES {
         if pairs.is_empty() {
-            return;
+            return Ok(());
         }
+        let conn = store.connection()?;
         let refs: Vec<(&str, &str)> = pairs
             .iter()
             .map(|(a, b)| (a.as_str(), b.as_str()))
@@ -206,7 +220,7 @@ pub(crate) fn copy_edges_with_bad_record_retry(
         match conn.query(&format!("COPY {table} FROM '{}'", fwd_slash_path(edge_pq))) {
             Ok(_) => {
                 let _ = std::fs::remove_file(edge_pq);
-                return;
+                return Ok(());
             }
             Err(e) => {
                 let msg = e.to_string();
@@ -227,12 +241,14 @@ pub(crate) fn copy_edges_with_bad_record_retry(
             }
         }
     }
+    let conn = store.connection()?;
     let refs: Vec<(&str, &str)> = pairs
         .iter()
         .map(|(a, b)| (a.as_str(), b.as_str()))
         .collect();
-    unwind_edges_from_pairs(conn, &refs, table, src_label, dst_label);
+    unwind_edges_from_pairs(&conn, &refs, table, src_label, dst_label);
     let _ = std::fs::remove_file(edge_pq);
+    Ok(())
 }
 
 pub fn classify_file(file: &str) -> &'static str {
--- a/crates/infigraph-core/src/resolve/calls.rs
+++ b/crates/infigraph-core/src/resolve/calls.rs
@@ -38,8 +38,8 @@ pub fn resolve_calls_incremental(
         symbol_map.entry(name).or_default().push((id, file, kind));
     }
 
-    let mut stats = resolve_with_map(&conn, extractions, &symbol_map, learned_store)?;
-    stats.inherits_resolved = resolve_inherits(&conn, extractions, &symbol_map)?;
+    let mut stats = resolve_with_map(store, &conn, extractions, &symbol_map, learned_store)?;
+    stats.inherits_resolved = resolve_inherits(store, extractions, &symbol_map)?;
     resolve_custom_edges(&conn, extractions, &symbol_map)?;
     Ok(stats)
 }
@@ -74,8 +74,8 @@ pub fn resolve_calls(
         }
     }
 
-    let mut stats = resolve_with_map(&conn, extractions, &symbol_map, learned_store)?;
-    stats.inherits_resolved = resolve_inherits(&conn, extractions, &symbol_map)?;
+    let mut stats = resolve_with_map(store, &conn, extractions, &symbol_map, learned_store)?;
+    stats.inherits_resolved = resolve_inherits(store, extractions, &symbol_map)?;
     resolve_custom_edges(&conn, extractions, &symbol_map)?;
     Ok(stats)
 }
@@ -230,6 +230,7 @@ fn write_external_calls(
 
 /// Caller must hold WriteLock.
 fn resolve_with_map(
+    store: &GraphStore,
     conn: &kuzu::Connection<'_>,
     extractions: &[FileExtraction],
     symbol_map: &HashMap<String, Vec<(String, String, String)>>,
@@ -607,7 +608,7 @@ fn resolve_with_map(
             .map(|(a, b)| (a.clone(), b.clone()))
             .collect();
         let pq_path = std::env::temp_dir().join("infigraph_resolve_calls.parquet");
-        copy_edges_with_bad_record_retry(conn, "CALLS", pairs, "Symbol", "Symbol", &pq_path);
+        copy_edges_with_bad_record_retry(store, "CALLS", pairs, "Symbol", "Symbol", &pq_path)?;
     }
 
     if !external_calls.is_empty() {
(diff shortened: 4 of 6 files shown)


murari316 (changes requested):

The PR fixes a real production bug: a caught bulk-COPY failure (bad-PK retry) leaves Kuzu's transaction bookkeeping on that connection wedged, so the next statement on the same connection fails with "Invalid transaction type to rollback". It threads &GraphStore into copy_edges_with_bad_record_retry and takes a fresh, cheap Connection per attempt, converting the function to return Result and adding fresh-connection points in the SCIP importer. The core change is sound and correctly scoped to write paths that already hold the WriteLock (no advisory-lock, kuzu-open-validation, or ignore-file invariants are affected), but CI is red on 2 of 3 platforms, the branch is unmergeable against current main (which refactored the CALLS write into write_resolved_calls, still using the shared-connection pattern this PR fixes), and no test covers the exact failed-COPY-then-next-COPY scenario the PR exists to fix.

---

### [BLOCKER] test-failure — `crates/infigraph-core/src/scip/mod.rs:236`
CI is red on ubuntu and windows (Check jobs failed; macOS passed; run is from Aug 25, ~3 weeks stale): scip::tests::scip_parameter_descriptor_does_not_become_a_new_symbol panics at scip/mod.rs:1069. Root cause is a shared fixed temp path: import_scip_index writes new-symbol batches to std::env::temp_dir().join("infigraph_scip_symbols.parquet") (scip/mod.rs:236) before COPY. Sibling tests in the same binary (e.g. non_implementation_relationship_does_not_create_inherits_edge, scip/mod.rs:885, which imports Animal/Dog through the same path) run in parallel and overwrite the file between write and COPY, so one test bulk-loads another test's rows into its database. Same class of bug the PR is about — a shared mutable resource reused across independent bulk COPY writes — and the same fixed paths exist at scip/mod.rs:481 (infigraph_scip_calls.parquet), scip/mod.rs:549 (infigraph_scip_inherits.parquet), resolve/calls.rs:610 (infigraph_resolve_calls.parquet), and resolve/inherits.rs:186 (infigraph_resolve_inherits.parquet). Outside tests, two concurrent indexing processes can cross-contaminate each other's graph DBs the same way. Fix: use unique_tmp_dir() as store_parquet.rs already does. Repo invariant requires the full suite green across macOS/Linux/Windows.

_Failure scenario:_ CI run 32899444920: assertion failed at scip/mod.rs:1069 — 'no node should exist for the parameter descriptor' — left: ["Animal", "Dog"], right: ["mintFn"]. stats.symbols_added==1 passes (counted in memory) but the DB contains the parallel inherits-test's Animal/Dog rows and no mintFn, exactly what a parquet overwrite between write_node_parquet and COPY produces.

### [MAJOR] stale-branch — `crates/infigraph-core/src/resolve/calls.rs:769`
The PR is unmergeable (GitHub reports mergeable: false) because main has since refactored the CALLS write out of resolve_with_map into write_resolved_calls (calls.rs:769-791 on main), which still calls copy_edges_with_bad_record_retry(conn, ...) on the shared connection and then reuses that same conn for write_external_calls. The PR's calls.rs hunks conflict structurally with main. On rebase the compiler will force a signature fix, but the author must thread &GraphStore into write_resolved_calls (and resolve_pairs callers) so the fresh-connection fix actually lands in main's newer code path — otherwise the exact wedge bug this PR fixes (Symbol COPY bad-PK retry wedging the connection used by the following CALLS COPY / external-calls write) survives.

_Failure scenario:_ Rebase onto current main: calls.rs hunks conflict; if resolved by keeping main's write_resolved_calls with its &Connection parameter and passing a shared conn through, a caught COPY CALLS failure inside write_resolved_calls wedges the connection that write_external_calls uses next, reproducing the production 'Invalid transaction type to rollback' failure on the new path.

### [MAJOR] test-coverage — `crates/infigraph-core/src/graph/store_util.rs:187`
No test covers the exact scenario the PR fixes: a caught bad-PK COPY failure followed by an independent bulk COPY on the same write path. The diff touches zero test files. The scip test module already has TestEnv infrastructure that makes this cheap: seed a Symbol table, call copy_edges_with_bad_record_retry with a batch containing an endpoint id that does not exist (forces the caught-failure retry path on one connection), then run a second table's COPY through the same path and assert its edges land. That directly encodes the production sequence (Symbol-table bad-PK retries → CALLS-table COPY) and would regress loudly if someone reintroduces a shared connection.

_Failure scenario:_ A future refactor reintroduces a long-lived shared connection into copy_edges_with_bad_record_retry (e.g. for perf, caching Connection::new). All existing tests pass because none exercises failed-COPY-then-next-COPY on one connection; the 'Invalid transaction type to rollback' wedge silently returns and is only caught in production.

### [MINOR] concurrency — `crates/infigraph-core/src/scip/mod.rs:380`
PLAUSIBLE (no direct evidence UNWIND failures wedge — only COPY is confirmed): the PR adds a fresh connection before the enrichment UNWIND loop (scip/mod.rs:380-383 at head) and before the Symbol UNWIND fallback loop (scip/mod.rs:349-352), but each loop then reuses that one connection across all chunks with errors swallowed via `let _ = conn.query(...)`. By the PR's own wedge model — 'a caught failure can leave the connection wedged for whatever query runs next on that same connection' — a failed chunk k can silently poison chunks k+1..N. Either take a fresh connection per chunk on failure, or stop swallowing the error.

_Failure scenario:_ During SCIP import of a large index, enrichment chunk 3 of 40 fails (e.g. transient Kuzu error) and wedges the shared connection; chunks 4-40's UNWIND SET queries all fail silently, leaving ~92% of docstring enrichments missing with no error surfaced — stats still report symbols_enriched as if they were written.

### [NIT] error-handling — `crates/infigraph-core/src/graph/store_util.rs:204`
Non-blocking: on the new early-error paths, the temp parquet file leaks. If store.connection()? fails on attempt N>0 (after write_edge_parquet already wrote edge_pq on a prior attempt), or before the UNWIND fallback (new `let conn = store.connection()?;` after the loop), the function returns Err without the `std::fs::remove_file(edge_pq)` cleanup that every other exit performs.

_Failure scenario:_ Connection creation fails under resource pressure mid-retry; infigraph_index_calls.parquet is left in the temp dir. Mostly cosmetic, but with the fixed-name temp paths from the blocker finding, a stale leftover file can also be picked up by a later COPY.

### [NIT] api-design — `crates/infigraph-core/src/graph/store_parquet.rs:196`
Non-blocking footgun, not a bug today: upsert_all_parquet_conn(&self, conn, ...) now performs node COPYs on the caller-supplied `conn` but edge COPYs on fresh connections from `self`. Verified the only caller (upsert_all_parquet, store_parquet.rs:192) passes self.connection(), so both target the same DB — but the signature now silently permits a caller to pass a connection to a different database and split nodes and edges across two DBs. Consider dropping the conn parameter (derive it from self) or documenting that conn must belong to self. Separately, the embed/mod.rs clippy allow is unrelated scope creep (CI's clippy accepts the lint name — macOS Check passed clippy — so it is not a CI risk, just worth a mention in the commit message).

_Failure scenario:_ A future multi-repo/combined-graph caller passes a combined-DB connection with a per-repo GraphStore as self: Module/File/Symbol nodes land in the combined DB while CALLS/INHERITS/etc. edges land in the per-repo DB, producing a graph with orphaned edges that fail bad-PK retries en masse.


pradeepmouli:
Pushed `3a0a7a8`, which fixes the BLOCKER, plus `46ad879`/`58298f9` rebasing this branch onto current `main`.

**The CI failure.** Root cause confirmed exactly as described. Five bulk loads staged through a fixed `temp_dir().join("infigraph_*.parquet")` — the SCIP symbol/calls/inherits imports and the CALLS/INHERITS resolve writes — with nothing serialising the window between writing the file and COPYing it. All five now use `unique_tmp_dir()`.

Reproduced it locally on macOS, despite macOS having passed in the CI run you cited — it is timing-dependent rather than platform-specific:

```
left: ["Animal", "Dog"]   right: ["mintFn"]
```

**1 failure in 4 runs before the change, 0 in 8 after.** The empirical figure is only corroboration though; the actual guarantee is structural, since the path is now keyed on pid plus a per-call counter and cannot collide by construction. I added a deterministic regression test alongside, so the guard doesn't depend on losing a race — the flaky scip test proves the symptom, the new test pins the invariant.

On the fix's shape: `unique_tmp_dir` already existed and was already doing exactly this job two functions away in `store_parquet.rs`, but was private to that module. It moves to `store_util.rs` — already `pub(crate)`, and already the import source for `copy_edges_with_bad_record_retry` at every one of these call sites — and `store_parquet` now uses the shared copy. One definition, five callers.

Worth stressing that this isn't only a test bug, as your note said: a daemon staleness reindex running alongside a CLI index is the normal configuration, and those two cross-contaminate each other's graph DBs through the same window.

**The stale-branch finding is resolved** by the rebase. I checked whether `a76a00a` had already subsumed this PR's fix, and it hasn't — `resolve_pairs` is now genuinely pure, but `write_resolved_calls` still passed one borrowed `Connection` to both `copy_edges_with_bad_record_retry` and `write_external_calls`, so the defect survives it. The fix now lands on `write_resolved_calls`, which takes `&GraphStore` and opens its own connection for the external-call write.

The unrelated `embed/mod.rs` clippy allow is now its own commit (`58298f9`), per your NIT — drop it if you'd rather fix that lint separately on `main`. I did check whether it was still needed: clippy 1.98.0 still fails on an unmodified checkout of current `main` without it.

Verification: 311/311 lib tests, `fmt` clean, `clippy -p infigraph-core -D warnings` clean. Not run: the repo's four `--ignored` perf gates — a full workspace build needs ~20 GB and I was working on a disk-constrained machine. Flagging rather than implying full coverage.

Remaining findings on this PR (the `store_util.rs:187` test-coverage MAJOR, and the MINOR/NIT items) are not addressed here.

PR #71, opened 2026-08-30
Fixes #73: invalidate the incremental cache on extractor change, not just file change

### LINKED ISSUE #73: Extractor fixes never reach existing indexes — content_hash only covers file bytes
`Module.content_hash` is a SHA-256 of the file's bytes and nothing else, and `index()` skips any file whose hash matches the stored one:

```rust
// crates/infigraph-core/src/extract/mod.rs
let content_hash = {
    let mut hasher = Sha256::new();
    hasher.update(source);
    format!("{:x}", hasher.finalize())
};
```

Nothing in that hash says anything about the extractor that parsed the file — no grammar version, no digest of the pack's `.scm` queries.

So when we fix an extractor, existing indexes never pick it up. 3.2.16 shipped `fix(java): capture method_reference (Foo::bar, this::bar) as call edges` and `fix(cpp): scope namespace-block functions to their namespace in symbol IDs`. If you upgrade without editing your source, those files still match their old hash and get skipped — not until the next release, but indefinitely.

Repro:

1. Index a Java project on a version before the `method_reference` fix
2. Upgrade, re-run `infigraph index` — no source edits, so every file is skipped
3. `trace_callers` on a method only ever called via `Foo::bar` still shows no callers, and it still turns up in `detect_dead_code`

`index --full` gives the right answer, which confirms the extractor itself is fine and only the skip decision is wrong.

The bad part is that it's silent — no warning, nothing indicating the index is behind the binary. Easy for someone to report a bug that's already been fixed.

Ideally the stored hash would cover the language pack too (query sources, grammar, plus an extractor version for Rust-side changes), so a Java query fix re-extracts Java files and leaves Python and TypeScript cached. Mixing in a global version would also work but would mean a full reindex on every release.

Same gap applies to the ANTLR grammar plugins — swapping a plugin's `.g4` or its extractor doesn't invalidate anything either.

learncoder4848:
## Summary

Fixes #73 .

`Module.content_hash` only covers the file's bytes, so `index()` skips any file whose contents haven't changed — even when the extractor that parsed it has. Extractor fixes therefore never reach an existing index. 3.2.16's Java `method_reference` and C++ namespace-scoping fixes both silently missed anyone who didn't happen to run `index --full`.

`LanguagePack` now carries a fingerprint built from its `entities.scm` / `relations.scm` sources, the grammar's identity, its custom edge definitions, and a new `EXTRACTOR_SCHEMA_VERSION` for Rust-side changes. `content_fingerprint` folds that into the per-file hash.

It's scoped per pack, so a Java query fix re-extracts Java files and leaves Python and TypeScript cached. Bumping a single global version would also work, but would mean a full reindex on every release.

Two places computed the file hash independently — the skip gate in `index_via_backend` and `extract_file`. Both now go through `content_fingerprint`. If those ever drift apart nothing gets skipped and every index quietly becomes a full one, so there's a test pinning it.

Grammar plugins get the same treatment via `plugin_fingerprint`, over the plugin's `plugin.toml`, its `.g4` grammars and its extractor source.

No schema migration — `content_hash` is already an opaque string in every backend.

One caveat: tree-sitter exposes no grammar version, so grammar identity is inferred from `abi_version` plus the node-kind / field / parse-state counts. An upgrade leaving all four unchanged would be missed. Rust-side extractor changes still need a manual `EXTRACTOR_SCHEMA_VERSION` bump. Both are documented in `docs/CODE-PARSING.md`.

## Test plan

- [x] `cargo test --all` — 1263 passed, 0 failed (13 new)
- [x] `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo check --features remote -p infigraph-core -p infigraph-cli -p infigraph-mcp` — clean (CI builds default features only; the release workflow uses `--features remote`)
- [x] `cargo test -p infigraph-core --test index_perf -- --ignored` — `test_incremental_index_skips_unchanged` still skips everything, confirming the skip gate and the stored hash agree

The main new one is `test_extractor_change_reindexes_unedited_files`: index a project, reopen with the same pack (nothing re-indexed), reopen with a changed pack (all re-indexed), reopen again with the changed pack (nothing). That last step is the one that matters — it proves this doesn't turn into a permanent reindex loop.

## Notes

Adding a private field to `LanguagePack` is technically breaking for struct-literal construction downstream, so this probably wants a minor bump rather than a patch.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -279,10 +279,10 @@ If the Model2Vec model files are not found, the embedder automatically falls bac
 
 ## 8. Incremental Indexing
 
-Every indexed file has its SHA-256 content hash stored in the `Module` node (`content_hash` property). On subsequent `infigraph index` runs:
+Every indexed file has a SHA-256 fingerprint stored in the `Module` node (`content_hash` property). The fingerprint covers both halves of staleness: the file's bytes **and** the language pack that parsed it (its `entities.scm`/`relations.scm` sources, the grammar's structural identity, its custom edge definitions, and `EXTRACTOR_SCHEMA_VERSION`). Upgrading an extractor therefore re-extracts the files that extractor owns without a `--full` reindex, and without disturbing other languages. On subsequent `infigraph index` runs:
 
-1. All files are hashed (in parallel via rayon)
-2. Files whose hash matches the stored hash are skipped entirely — no re-parsing, no graph updates
+1. All files are fingerprinted (in parallel via rayon) via `LanguagePack::content_fingerprint`
+2. Files whose fingerprint matches the stored one are skipped entirely — no re-parsing, no graph updates
 3. Changed and new files are re-parsed and their nodes/edges are deleted and reinserted
 4. The cross-file call resolution pass only re-resolves calls from changed files, but reads the full symbol table from the graph (so cross-file edges from unchanged files are preserved)
 
--- a/crates/infigraph-core/src/extract/mod.rs
+++ b/crates/infigraph-core/src/extract/mod.rs
@@ -4,7 +4,6 @@ pub use entities::extract_entities;
 pub use relations::{extract_relations, extract_relations_with_custom_edges};
 
 use anyhow::Result;
-use sha2::{Digest, Sha256};
 
 use crate::analysis::extract_statements;
 use crate::lang::{LanguagePack, ParserBackend};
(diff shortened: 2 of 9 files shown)


learncoder4848:
@murari316 @johnintuit @sandeep-mewara @WinterQuant Can you guys review the changes ?

learncoder4848:
@murari316 @johnintuit @sandeep-mewara @WinterQuant @pradeepmouli 

**Reminder** 
Can you guys review the changes and start the workflow ?

learncoder4848:
Reminder @murari316 @johnintuit @sandeep-mewara @WinterQuant @pradeepmouli

PR #74, opened 2026-09-01
fix(core): disambiguate genuinely-new SCIP symbols sharing a (file, name) key

pradeepmouli:
## Depends on #72

This PR continues directly from #72 (\`fix(core): scope Rust impl methods to their type, dedup class-scoping walk\`), which is still open. #72's own description tracks the remaining work as pradeepmouli/infigraph#125 and #126 — this PR is that work, plus one more fix on top.

**This branch includes #72's commit** (`7f05b59` there, re-applied here as its own commit of identical content) purely so the branch builds and its tests pass in isolation. Once #72 merges, this should be rebased onto the then-current `main` — the diff will shrink to just the commits below, with no duplication. Please don't merge this before #72, or hold off reviewing the first commit specifically (it's a re-run of #72's own diff, not new).

## Summary

The symbol-identity-and-scoping-hardening spec's Phase 2 (Parts A & B) added ordinal disambiguation for genuinely-distinct symbols that share a derived id — but left one residual gap: SCIP enrichment's `.first()` fallback (a deliberate narrow escape hatch for "a genuine span-computation edge case against a single pre-existing symbol") fires far too broadly, because `file_name_to_ids` is mutated *during* the same Pass 1 loop that reads it. A second brand-new same-named symbol in the same import batch sees the first one (inserted moments earlier, same loop) as its only candidate, fails span-containment, and silently merges into it as an enrichment instead of becoming its own symbol.

Found via a real production incident: sittir's daemon repeatedly hit "duplicated primary key" COPY failures during SCIP enrichment, traced to two distinct same-named symbols (two same-named methods on different types) colliding into one id.

Fix, in two parts:
- Scope the `.first()` fallback to a snapshot of what existed in the graph *before* this import pass, never against a candidate gained during the same Pass 1 loop, and only when pre-existing ambiguity is a single candidate.
- Apply the same ordinal-disambiguation scheme already used by the tree-sitter path (`entities.rs`, Phase 2 Part A) to genuinely-new same-named symbols, instead of relying on the `seen_ids` dedup, which silently kept only one and dropped the rest. Ordinals continue past whatever count already existed for that key, so a new symbol can never collide with one already in the graph.

Also included (found necessary during independent verification of this branch against a fresh checkout of `main`):
- `fix(scip): propagate preload query failures instead of silently swallowing them` — a pre-existing gap in the same file that two of this branch's own regression tests caught when run against a fresh base; without it, `import_scip_index` silently swallows a failed preload query instead of surfacing it.
- A one-line, unrelated clippy fix (`chunks_exact` → `as_chunks`) in `embed/mod.rs`, needed only to satisfy this repo's pre-commit hook on the base commit — zero functional change.

Two new regression tests: two brand-new same-named symbols in one file both survive with distinct ordinal ids; a new symbol colliding with two pre-existing same-named symbols continues the ordinal sequence instead of colliding with either.

## Verification

Independently verified on a fresh branch off `upstream/main` (not just cherry-picked and trusted): full workspace build, `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and the full `infigraph-core` lib suite (318 tests) all green.

https://claude.ai/code/session_01HdmNge4878WaicQ4QPfZme

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2495,6 +2495,7 @@ dependencies = [
  "tree-sitter-dart",
  "tree-sitter-kotlin-ng",
  "tree-sitter-python",
+ "tree-sitter-rust",
  "ureq 2.12.1",
  "usearch",
 ]
--- a/crates/infigraph-core/Cargo.toml
+++ b/crates/infigraph-core/Cargo.toml
@@ -69,6 +69,7 @@ optional = true
 tree-sitter-python = "0.25"
 tree-sitter-kotlin-ng = "1.1"
 tree-sitter-dart = "0.2"
+tree-sitter-rust = "0.24"
 cargo-husky = { version = "1", default-features = false, features = ["user-hooks"] }
 # Dev-dep cycle (infigraph-languages -> infigraph-core) is fine: dev-deps are excluded
 # from the normal build graph. Needed so remote_cross_service.rs can index real
--- a/crates/infigraph-core/src/embed/mod.rs
+++ b/crates/infigraph-core/src/embed/mod.rs
@@ -399,8 +399,10 @@ pub fn load_embeddings(path: &Path) -> Result<Vec<(String, Vec<f32>)>> {
         let float_bytes = dim * 4;
         anyhow::ensure!(pos + float_bytes <= data.len(), "truncated embeddings file");
         let vec: Vec<f32> = data[pos..pos + float_bytes]
-            .chunks_exact(4)
-            .map(|chunk| f32::from_le_bytes(chunk.try_into().unwrap()))
+            .as_chunks::<4>()
+            .0
+            .iter()
+            .map(|chunk| f32::from_le_bytes(*chunk))
             .collect();
         pos += float_bytes;
         result.push((id, vec));
--- a/crates/infigraph-core/src/extract/entities.rs
+++ b/crates/infigraph-core/src/extract/entities.rs
@@ -1,6 +1,7 @@
 use sha2::{Digest, Sha256};
 use tree_sitter::{Node, Query, QueryCursor, StreamingIterator};
 
+use super::{find_parent_class, node_text, resolve_compound_node_text};
 use crate::analysis::cyclomatic_complexity;
 use crate::model::{Span, Symbol, SymbolKind};
 
@@ -18,12 +19,19 @@ use crate::model::{Span, Symbol, SymbolKind};
 ///   @test.def / @test.name / @test.docstring
 ///   @var.def / @var.name
 ///   @route.def / @route.method / @route.path / @route.handler
+///   @method.parent (optional -- only needed when a grammar's enclosing node has
+///     no generic "name" field to walk up to, e.g. Rust's `impl_item`; see Rust's
+///     entities.scm. Preferred over the generic ancestor walk when present.
+///     `decompose_query`, when present, resolves a compound `@method.parent`
+///     capture (e.g. a generic `impl<T> Foo<T> for Vec<Bar>`'s `type:` field)
+///     down to its base identifier -- see `resolve_compound_node_text`.)
 pub fn extract_entities(
     file: &str,
     source: &[u8],
     root: Node,
     query: &Query,
     language: &str,
+    decompose_query: Option<&Query>,
 ) -> Vec<Symbol> {
     let mut cursor = QueryCursor::new();
     let mut matches = cursor.matches(query, root, source);
@@ -44,6 +52,7 @@ pub fn extract_entities(
         let mut route_path: Option<String> = None;
         let mut route_handler: Option<String> = None;
         let mut route_def_node: Option<Node> = None;
+        let mut parent_capture: Option<Node> = None;
 
         for capture in m.captures {
             let idx = capture.index as usize;
@@ -84,6 +93,9 @@ pub fn extract_entities(
                 }
                 "method.decorator" => {
                     decorator = Some(node_text(node, source));
+                }
+                "method.parent" => {
+                    parent_capture = Some(node);
                 }
                 "func.params" | "method.params" => {
                     params_capture = Some(node_text(node, source));
@@ -206,8 +218,12 @@ pub fn extract_entities(
                 _ => hash_node(node, source),
             };
 
-            // Find parent class for methods by walking up the AST
-            let parent_class = find_parent_class(node, source);
+            // Prefer an explicit @method.parent/@func.parent capture (needed when a
+            // grammar's enclosing node has no generic "name" field to walk up to,
+            // e.g. Rust's impl_item) over the generic ancestor walk.
+            let parent_class = parent_capture
+                .map(|n| resolve_compound_node_text(n, source, decompose_query))
+                .or_else(|| find_parent_class(node, source));
             let id = if let Some(ref cls) = parent_class {
                 format!("{}::{}::{}", file, cls, name)
             } else {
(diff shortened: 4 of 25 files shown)


murari316 (changes requested):

The PR disambiguates same-(file,name) symbols in both the tree-sitter path (ordinal-suffixed ids in entities.rs) and the SCIP import (span-containment matching, moniker-keyed edge resolution, a scip_id column, and loud preload failures), with substantial regression tests. The design is sound and well-documented, but the core containment mechanism compares SCIP's 0-based occurrence lines against tree-sitter's 1-based stored spans, so it fails in the dominant first-line-identifier case — masked by the single-candidate fallback on typical repos and by tests that hand-craft rows in the 0-based convention — and the ordinal rename leaves stale ids in the Pass-2 caller-attribution map. Confidence: High — every claim traced to the diff plus entities.rs L196 (row+1) and parse_range L560-573 (no conversion) on main. Recommendation: hold for fixes.

---

### [BLOCKER] correctness — `crates/infigraph-core/src/scip/mod.rs:246`
The new containment match compares a SCIP occurrence's 0-based start_line (parse_range, L560-573, passes SCIP lines through unchanged) against tree-sitter symbol spans stored 1-based (entities.rs L196/L263: `row + 1`), so containment is off by one and fails whenever the definition identifier sits on the symbol's first line — the dominant case (`def foo():`, `fn foo(`). Single-candidate keys are rescued by the `.first()` fallback, so typical repos look fine; but for multi-candidate keys — the exact scenario this PR exists to fix — the fallback refuses (`c.len() != 1`), the occurrence takes the new-symbol path, and a duplicate Symbol row is inserted for an already-indexed symbol. That is a regression vs. main, which over-enriched but never duplicated. Note for the fix: parse_range output is also the stored span of SCIP-created symbols, so the graph holds mixed-base rows (tree-sitter 1-based, SCIP-created 0-based); a naive +1 at the comparison site alone breaks re-import containment against previously stored SCIP rows.

_Failure scenario:_ File has `impl A { fn foo... }` (tree-sitter span [10,15], id a::A::foo) and `impl B { fn foo... }` ([20,25], a::B::foo). SCIP emits definition occurrences at 0-based lines 9 and 19. Containment: 9>=10 false, 19>=20 false; fallback sees 2 pre-existing candidates and refuses; both defs are treated as brand-new, ordinal-suffixed rows `a.rs::foo#3`/`a.rs::foo#4` are COPY-inserted alongside the two real symbols — duplicate nodes, and neither real symbol gets its docstring/scip_id enrichment.

### [MAJOR] test-coverage — `crates/infigraph-core/src/scip/mod.rs:1420`
The new tests encode the same off-by-one as the code instead of catching it: they hand-CREATE Symbol rows whose start_line values match the raw 0-based SCIP occurrence lines (e.g. `calls_edge_resolves_to_the_specific...` creates A::foo with start_line 1 and passes `range: vec![1,4,7]`, with a comment claiming the parsed value is '1-based line 2' — parse_range performs no such conversion). No test runs real tree-sitter extraction followed by a real SCIP import, which is what would expose the base mismatch; the repo convention of extending tests/fixtures/microservices (which exercise the full extract→import pipeline) would have caught it.

_Failure scenario:_ All new tests pass while the shipped containment logic never matches a genuinely tree-sitter-extracted symbol whose identifier is on its span's first line; the bug reaches production undetected.

### [MAJOR] correctness — `crates/infigraph-core/src/scip/mod.rs:322`
The ordinal-rename block mutates only `new_symbols[i].0` and `scip_sym_to_ts_id`; the `file_symbols` (and `file_name_to_ids`) entries pushed for those symbols during Pass 1 keep the stale unsuffixed id. Pass 2's caller attribution (unchanged in this PR) resolves `container_id` from `file_symbols` by containment, producing a CALLS edge FROM a node id that no longer exists after the rename; `copy_edges_with_bad_record_retry` then drops it as a bad-PK record with no warning. Because new SCIP symbols' stored spans are the identifier token (~1 line), this fires when a reference shares the definition's line — e.g. one-line functions/arrow functions, which are common in TS/JS — and finding #1 inflates how many symbols take the new-symbol path in the first place. A sibling symptom: a second in-batch definition occurrence that containment-matches a stale candidate pushes an enrichment row targeting the dead unsuffixed id (silent no-op UNWIND).

_Failure scenario:_ index.scip contains new same-named one-line symbols `const foo = () => bar()` in classes A and B; both are renamed to test.ts::foo#1/#2, but file_symbols still says test.ts::foo. The reference to bar on foo#1's line attributes container_id = 'test.ts::foo' (nonexistent) → COPY CALLS fails on that record → the edge is silently dropped.

### [MINOR] correctness — `crates/infigraph-core/src/scip/mod.rs:265`
The `.first()` fallback's `c.len() == 1` guard counts pre-existing candidates but not how many distinct monikers use the fallback, so any number of genuinely-new same-named symbols whose spans fail containment all merge into the single lone pre-existing symbol — the same silent-collapse failure mode the PR fixes for the no-preexisting case. Each pushes an enrichment (last scip_id wins) and maps its moniker to the same ts id, misdirecting Pass 2/3 edges.

_Failure scenario:_ Graph has one tree-sitter `foo` at [1,5] (e.g. macro-heavy Rust file where tree-sitter found only one of three impls). SCIP defines A#foo (line 2), B#foo (line 10), C#foo (line 20). B and C fail containment, both hit the single-candidate fallback, both 'enrich' the [1,5] symbol; no new rows are created and all references to B::foo/C::foo resolve to the wrong node.

### [MINOR] correctness — `crates/infigraph-core/src/scip/mod.rs:345`
Ordinal continuation uses `existing_count` (candidate count) rather than the max existing `#n` suffix, so a gapped ordinal set — e.g. `foo#2`,`foo#3` remaining after a prior run's bad-PK record drop or a partial delete — makes the new id `file::foo#3` collide with a pre-existing row; the COPY retry loop then drops the new symbol silently.

_Failure scenario:_ Prior state: test.ts::foo#2 and test.ts::foo#3 exist (foo#1 was dropped by an earlier COPY bad-record retry). New D#foo arrives; existing_count=2 → id test.ts::foo#3 → duplicated primary key → extract_bad_copy_value drops it → symbol lost with only an eprintln.

### [MINOR] correctness — `crates/infigraph-core/src/scip/mod.rs:430`
The learned-correction detection derives `call_name` via `target_id.rsplit("::").next()`, which now yields ordinal-suffixed names like `x#2` for disambiguated targets; the comparison against tree-sitter targets' names (which may carry a different or no suffix) can no longer match, so corrections involving disambiguated symbols are silently never learned (or spuriously learned).

_Failure scenario:_ SCIP resolves a call to test.ts::Bar::x#2 while tree-sitter had an edge to test.ts::Bar::x#1; call_name 'x#2' != 'x#1' so `ts_had_different` is false and the correction is not recorded.

### [MINOR] consistency — `crates/infigraph-core/src/graph/cozo_store.rs:713`
CozoStore::import_symbols adds the scip_id column but always writes a hardcoded empty string (the 14-tuple input carries no scip_id), so any migration/export through the Cozo path silently drops SCIP enrichment data that the kuzu and neo4j backends persist.

_Failure scenario:_ A project with SCIP-enriched symbols is migrated via import_symbols into Cozo; every row's scip_id comes back '' and moniker-based lookups against the Cozo backend find nothing.

### [NIT] style — `crates/infigraph-core/src/extract/entities.rs:1806`
Contradictory comments on the rust_entity_query test helper: the test doc comment above it claims it 'Uses the real bundled Rust language pack (entities.scm + inherit_decompose_query), not a hand-rolled query', while the helper's own comment says it 'hand-builds the query rather than loading the registry' and is 'kept in sync manually'. One of the two must be corrected so future readers know the queries can drift from the shipped .scm files.

_Failure scenario:_ rust/entities.scm changes its impl-method pattern; the test keeps passing against the stale inline copy while readers of the doc comment believe the shipped query is exercised end-to-end.

### [NIT] scope — `crates/infigraph-core/src/embed/mod.rs:399`
The chunks_exact -> as_chunks::<4>() refactor in load_embeddings is unrelated to this PR's purpose and silently raises the effective MSRV to Rust 1.88 (slice::as_chunks stabilization); since CI floats on dtolnay stable this passes today, but it belongs in its own change, not a correctness PR.

_Failure scenario:_ A contributor or downstream consumer pinned below 1.88 gets a compile error in a PR ostensibly about SCIP symbol disambiguation.


pradeepmouli:
Pushed `fa18cfd`, which fixes the BLOCKER.

Confirmed exactly as diagnosed: `parse_range` passed SCIP's 0-based lines through unchanged, and they were compared against spans stored 1-based by tree-sitter (`entities.rs` writes `row + 1`).

**Fixed at the boundary, not the comparison** — taking your note seriously that a naive `+1` at the comparison site breaks re-import. `parse_range` now normalises, so both the containment check *and* the stored span of a SCIP-created symbol are 1-based. A comparison-site fix would have left the graph holding mixed-base rows and the next import hitting the same mismatch from the other direction.

**A second site, which isn't in the review:** the references pass computes `ref_line` from `occ.range` directly and compares it against the same `file_symbols` spans, so container attribution for references had the identical off-by-one. Both entry points now go through one `scip_line_to_1based`.

**Regression test** covers the multi-candidate case the `.first()` fallback cannot rescue — two same-named methods with distinct spans, definition occurrences on each one's first line:

```
before: 4 rows, symbols_added=2   (two duplicates)
after:  2 rows, symbols_added=0, symbols_enriched=2
```

**One existing test changed, and I want to flag it rather than bury it.** `two_new_same_named_symbols_in_one_file_both_survive_with_distinct_ordinal_ids` now expects `start_line` 2 and 11 instead of 1 and 10. Its fixture emits occurrences at 0-based lines 1 and 10, and SCIP-created symbols are now stored 1-based like every other row in the graph — so the old expectation was encoding the mixed-base storage this commit removes. What that test actually pins, two distinct ordinal-suffixed ids ordered by span position, is unchanged.

**Migration note:** a graph built before this change may hold SCIP-created rows stored 0-based, and those won't match containment until the project is reindexed. Tree-sitter rows — the overwhelming majority, and the ones this bug actually broke — are correct immediately.

Verification: 319/319 lib tests, `fmt` clean, `clippy -p infigraph-core -D warnings` clean. Not run: the repo's four `--ignored` perf gates, since a full workspace build needs ~20 GB and I was disk-constrained. Flagging rather than implying full coverage.

The remaining findings on this PR — the `scip/mod.rs:1420` test-coverage MAJOR, the `scip/mod.rs:322` correctness MAJOR, and the MINOR/NIT items — are not addressed here.

PR #77, opened 2026-09-13
fix(mcp): bind Web UI/API server to loopback by default (ADV-2598)

murari316:
**Fixes ADV-2598** — P1 security defect: unauthenticated remote code execution via the Web UI `/api/query` endpoint.

## Problem
`infigraph-mcp --mcp` / `--ui` starts the Web UI/API server bound to `0.0.0.0:9749` (all interfaces). Because `--mcp` — the command every configured coding agent launches — sets `ui_enabled = true`, a normal stdio-MCP install exposes this server to the whole local network.

The `/api/query` route forwards a raw `cypher` string straight to the embedded Kuzu/LadybugDB engine with no validation. `check_auth()` is never invoked on the port-9749 server (only on the port-8642 Remote MCP server), and it is a no-op on default installs anyway — it returns `true` when `INFIGRAPH_API_KEY` is unset. An unauthenticated peer on the same network can therefore issue `FORCE INSTALL <name> FROM '<url>'` + `LOAD <name>` to `dlopen()` an attacker-controlled native library and execute code as the developer.

## Fix
- Bind `start_ui_server` to `127.0.0.1` by default, matching the documented `localhost:9749` intent (`ARCHITECTURE.md`, `CHANGELOG.md`).
- Add an `INFIGRAPH_UI_BIND` env var to opt into a wider bind (e.g. `0.0.0.0`) when the UI is intentionally exposed.
- Add a regression test (`test_ui_server_binds_loopback_not_wildcard_by_default`) asserting the wildcard address stays free by default.

Note: the ticket's suggested "invoke `check_auth()`" remediation is a no-op on default installs (no API key set), so the loopback bind is the load-bearing fix — and is exactly what the ticket's own verification step checks.

## Verification (ticket's Remediation Verification steps)
- `lsof` on `:9749` shows **`127.0.0.1`-only** for the patched build (unpatched build binds `*:9749`).
- Unauthenticated `POST /api/query` over the LAN IP is **connection-refused** against the patched build (the unpatched build returned `HTTP 200`). Tested with a benign `MATCH (n) RETURN count(n)` query, not the extension-load PoC.
- Loopback `POST /api/query` still returns `200` for the local user.
- `INFIGRAPH_UI_BIND=0.0.0.0` restores the wildcard bind for intentional exposure.
- `cargo fmt --check` and `cargo clippy -D warnings` clean; all web module tests pass.


murari316:
Closing in favor of a clean single-commit PR from a fresh branch (history got tangled by merge commits during force-pushes). The ADV-2598 fix is unchanged.

